Issue #20, opened 2017-08-10
Can not act on element position within ScrollTrack


Eldwick:
The children of `ScrollTrack` do not have all the necessary information to determine it's current position within the scroll track. 

I suggest we pass down the current state of scrollTrack and let the children determine how to consume that information.

PR #21, opened 2017-08-10
Add equalWidthTrack wrapper

Eldwick:
In regards to #20 

Pass state and nodeWidths to direct children of ScrollTrack.


src/components/ScrollTrack/ScrollTrack.js
```diff
@@ -56,6 +59,14 @@ class ScrollTrack extends Component {
     window.removeEventListener('resize', this.debouncdComputeSlideAttributes)
   }
 
+  get childrenWithTrackProps() {
+    const nodeWidths = this.getNodeWidths()
+
+    return React.Children.map(this.props.children, child => {
+      return React.cloneElement(child, { ...this.state, ...nodeWidths })
+    })
+  }
```
dcocchia: Should we namespace these? They're somewhat generically named within the component, and may collide with the children's own props.

Eldwick: 👍 good point

src/components/ScrollTrack/equalWidthTrack.js
```diff
@@ -0,0 +1,33 @@
+import React, { PureComponent } from 'react'
```
dcocchia: Can we add a comment block to the top of this file to describe it a bit more? When/why to use it, for example. We should add similar documentation in the ScrollTrack markdown file.

Eldwick: Will do

PR #25, opened 2017-08-11
Allow absolute path Fixes #24

### LINKED ISSUE #24: Allow absolute paths

nbwar:
Add the ability to use absolute path and changed components that used relative paths to absolute.
Also cleaned up some minor indenting and sorting of imports.


src/components/Buttons/CircleButton.js
```diff
@@ -3,11 +3,11 @@
  *
  * @author [Dominic Cocchiarella ](https://github.com/docchia)
 */
-import colors       from '../../styles/colors'
-import responsive   from '../../styles/responsive'
-import themer       from '../../styles/themer'
-import zIndex       from '../../styles/zIndex'
-import { spacings } from '../../styles/spacing'
+import colors       from 'styles/colors'
```
Eldwick: Kinda feels like we should import every style related things in one line, eg: 

```
import { colors, responsive, themer, zIndex, spacing } from 'styles'
```

src/components/Buttons/CircleButton.js
```diff
@@ -3,11 +3,11 @@
...
+import responsive   from 'styles/responsive'
+import themer       from 'styles/themer'
+import zIndex       from 'styles/zIndex'
+import { spacings } from 'styles/spacing'
```
Eldwick: If we do want to continue importing directly from the specific directory, this should mirror other style import conventions and be:

```
import spacings from 'styles/spacing'
```

Eldwick:
Fixes #24

src/components/Buttons/CircleButton.js
nbwar: I agree, but this should probably be a separate issue all I did was change from a relative path to an absolute path

dcocchia:
This looks good to me, but will need conflicts resolved. @nbwar can you resolve conflicts and then I can approve/merge?

PR #45, opened 2017-09-15
add appropriate roles for tab and tablist on navigation pills in store chooser

austinbourdier:
As it stands, the navigation pills in the store chooser don't have appropriate roles assigned to them.


dcocchia:
@austinbourdier Snapshot tests need to be updated for tests to pass. `jest test --u` will update snapshots to the current React components' outputs.

dcocchia:
Not all uses of NavigationPill will be as a tab. I think we can do this by allowing elementAttributes to be passed as a separate prop.

```js
<NavigationPill
  elementAttributes={{ role: 'tab', ariaLabel='this tab' }}
/>
```

would output
```
<li role='tab' aria-label='this tab'>...</li>
```

We'll be doing something very similar for the Button component, so would be consistent as well.

dcocchia:
Closing this in favor of https://github.com/instacart/Snacks/pull/59

PR #48, opened 2017-09-26
Themer updates

dcocchia:
### What
- Adds new `withTheme` HOC
- Adds two new attributes to theme config colors
- Update snacks components currently using the `themer`
- Test and docs updated

### Why
The `themer` alone does not work well with React components when the theme is changed. React components must know when a theme changes so that they can re-render. The HOC re-renders for you and passes down themes through props. The `themer` is still exported so that users can use Snacks without being tied to React and can still update the theme.

New colors are to match new colors needs for PBI sites.

### How
```js
import React, { Component } from 'react'
import { withTheme } from 'ic-snacks'

@withTheme()
class MyComponent extends Component {
  render() {
    <p style={{ color: this.props.snacksTheme.color.action }}>
      Hi!
    </p>
  }
}

export default MyComponent
```


src/styles/themer/withTheme.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    onThemeChange() {
+      this.forceUpdate()
+    }
+
+    render() {
+      return (
+        <InnerComponent
+          ref={(node) => this.wrapped = node}
```
Eldwick: Why do we need the ref here?

src/styles/themer/withTheme.js
```diff
@@ -0,0 +1,41 @@
...
+          snacksTheme={themer.themeConfig}
```
Eldwick: We might want to put this after we spread `{...this.props}`, someone may accidentally pass `snacksTheme` and overwrite the prop.

src/styles/themer/withTheme.js
```diff
@@ -0,0 +1,41 @@
...
+
+    componentDidMount() {
+      this.unsubscribe = themer.subscribe(this.onThemeChange)
+    }
+
+    componentWillUnmount() {
+      this.unsubscribe()
+    }
+
+    onThemeChange() {
```
Eldwick: seems like this may need to be defined as an attribute i.e `onThemeChange = () =>` since it's being passed as a callback

src/styles/themer/docs/withTheme.md
```diff
@@ -0,0 +1,33 @@
+The withTheme higher order component is the best way to use Snacks themes in your own React components. It tracks updates to your theme config and re-renders your components as needed. Theme attributes are passed to your components via a `snacksTheme` prop. Using this or the themer directly allows you to share a theme across snacks components and your own components.
```
wfro: Is there any reason to use `themer` directly?

src/styles/themer/docs/withTheme.md
```diff
@@ -0,0 +1,33 @@
...
+
+Class with decorator style
+```js static
+  import React, { Component } from 'react'
+  import { withTheme } from 'ic-snacks'
+
+  @withTheme()
```
wfro: This requires `themer` to be configured in whatever app is importing from snacks right? Might be worth a mention

src/styles/themer/index.js
```diff
@@ -16,6 +21,7 @@ class Themer {
 
   set themeConfig(themeConfig) {
```
wfro: Nit and getting mostly into personal preference, but especially given that `themer.themeConfig = { ... }` now triggers a callback as well, I'm less of a fan of the `set themeConfig` api. Triggering a callback on an assignment might be confusing. Don't necessarily feel strongly enough to change it though

src/styles/themer/withTheme.js
wfro: +1 , unused as is, but maybe in prep for hoisting statics?

wfro:
left a few random comments but LGTM

src/styles/themer/docs/withTheme.md
dcocchia: If someone wants to use Snacks along with a non-react library or to change the theme config.

src/styles/themer/withTheme.js
dcocchia: Yeah, I couldn't decide if that should be allowed or disallowed. I went with allowing it to be overwritten for more flexibility. What do you think @wfro?

src/styles/themer/docs/withTheme.md
dcocchia: good point 👍

src/styles/themer/index.js
dcocchia: I think it works pretty well. The other way this could work would an older style set
```js
themer.setConfig(config)
```

I personally like the setter because this way there's no way to muck with the internal theme even if you really want to.

src/styles/themer/withTheme.js
dcocchia: oops, yep. Was going to do hoisting and decided against it. Will remove for now.

src/styles/themer/withTheme.js
wfro: Don't feel too strongly either way but I'm find with it being overridable, most won't ever interact with the `snacksTheme` prop directly anyway, and so doesn't seem like a major footgun

dcocchia:
@Eldwick @wfro PR updated - ready for approval I think

src/styles/themer/withTheme.js
Eldwick: But since this is all the HOC does, seems weird to allow the only prop it passes down to be overridden

dcocchia: Hm, that's a really good point. 😓  Why override it if you're using the HOC?

Ok, going to change the ordering.

PR #52, opened 2017-09-29
Add <Button />

wfro:
Adds a `Button` component and related docs.

`Button.md` is the best to get an overview of what `Button` does.


src/components/Buttons/Button.js
```diff
@@ -0,0 +1,266 @@
...
+
+Button.propTypes = {
+  /** Size of the button. */
+  size: PropTypes.oneOf(['tiny', 'small', 'standard', 'large']),
+
+  /** Snacks button variants. */
+  snacksStyle: PropTypes.oneOf(['primary', 'secondary', 'coupon']),
+
+  /** Optional style overrides. */
+  styles: PropTypes.object,
```
dcocchia: I think we agreed to use `style` singular in Snacks to make it feel more like a regular style attribute.

Our other components use variations of both, so we'll need to update those some day :D

src/components/Buttons/Button.js
```diff
@@ -0,0 +1,266 @@
...
+    If href is provided, `Button` will render as an anchor
+    tag with the supplied href value.
+  */
+  href: PropTypes.string,
+
+  /** Text content for the button. */
+  children: PropTypes.node,
+
+  /** Any additonal props to add to the element (e.g. data attributes). */
+  extraProps: PropTypes.object,
```
dcocchia: I think this name is a bit confusing. They're almost explicitly not props, so I think calling something like `elementAttributes` or something would be more clear

src/components/Buttons/__tests__/Button.spec.js
```diff
@@ -0,0 +1,137 @@
...
+    const tree = renderer.create(
+      <StyleRoot>
+        <Button href="/carrot" snacksType="primary">
+          Hi
+        </Button>
+      </StyleRoot>
+    )
+    expect(tree).toMatchSnapshot()
+  })
+})
```
dcocchia: should add a test for making sure the extraProps/elementAttributes get added to the outputted element

src/components/Buttons/docs/Button.md
```diff
@@ -0,0 +1,180 @@
...
+    <Button snacksStyle="secondary" size="large">
+      Secondary Button
+    </Button>
+  </div>
+</div>
+```
+
+### Coupon
+
+Note that coupon buttons aren't affected by a Snacks theme.
```
dcocchia: Did we confirm with retailer team that this is what we want?

src/components/Buttons/Button.js
```diff
@@ -0,0 +1,266 @@
...
+    style: [
+      baseStyles,
+      sizeStyles[props.size],
+      snacksStyles[snacksStyle].base,
+      props.inverted && snacksStyles[snacksStyle].inverted,
+      ElementType === 'a' && linkStyles,
+      props.styles
+    ],
+    onClick: e => {
+      e.preventDefault()
```
Eldwick: We might not want to always preventDefault on all buttons

src/components/Buttons/Button.js
```diff
@@ -0,0 +1,266 @@
...
+      props.onClick()
```
Eldwick: We should think about standarizing what we pass to prop event handlers. I really like the way that react-semantic-ui does it. For all event handlers they pass two arguments, the event and all the props of the component: 
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Button/Button.js

That way the handler in the component can decide what to do with the event and opens up cooler things you can do with composition, allowing a handler defined higher up in the tree to make decisions based on the current props of the button being pressed.

dcocchia: ➕ ➕

src/components/Buttons/Button.js
```diff
@@ -0,0 +1,266 @@
...
+    },
+    ...props.extraProps
+  }
+  if (props.href) {
+    finalProps.href = props.href
+  }
+
+  let icon = props.icon
```
Eldwick: Yesterday, dom was talking about how we want to prefer const to let:

```
const icon = typeof props.icon === 'string' ? <Icon name={props.icon} /> : props.icon
```

src/components/Buttons/Button.js
wfro: I can get behind this

src/components/Buttons/Button.js
wfro: Funny enough, i actually originally called it `elementAttributes` and then changed it :D

src/components/Buttons/Button.js
wfro: @Eldwick yeah good point, should probably pass along the event object as well

src/components/Buttons/Button.js
wfro: Yeah I like this as well

src/components/Buttons/Button.js
wfro: I went back and forth between the two, I felt like this was more readable but I'm like 51/49, so I'll go with whatever the group wants @Eldwick @dcocchia

dcocchia: We should only prefer it when its unnecessary to use `let` or `var`. I think using `let` to increase readability, or for actual reassignments, is ok. Of course that's a bit subjective :D

src/components/Buttons/docs/Button.md
wfro: @dcocchia checking now, I'm not sure which theme color we'd want to use here. We'd likely need to add an option, even the default IC theme wouldn't have something to cover coupons.

wfro:
@dcocchia @Eldwick A few others notes -- the design system buttons have three colors for each button state (normal/hover/pressed), but snacks themes only cover normal/hover. It makes the buttons feel much less dynamic without feedback while pressing, so curious if anyone has any thoughts. Finding/writing some code to darken a hex string seems plausible

src/components/Buttons/docs/Button.md
wfro: confirmed that coupons aren't themable

dcocchia:
@wfro Yeah I noticed this with the circular button when I moved it to themed. I bet we can come up with a way to do this programmatically. I @jordanscode wanted to do it that way anyway - something like 10% darker on :active state (clicked).

wfro:
@dcocchia this seems like it would do the trick and is relatively active https://github.com/Qix-/color, or we could roll our own. any preference?

dcocchia:
@wfro Let's find out what the requirements are from @jordanscode first and go from there. I'm generally opposed to bringing in other libs with dozens of features only to use one feature -- but I'm not dogmatically against it.

src/components/Buttons/Button.js
```diff
@@ -0,0 +1,290 @@
...
+  display: 'flex',
+  alignItems: 'center',
+  boxSizing: 'border-box'
+}
+
+const getSnacksStyles = props => {
+  const { action, actionHover, primaryBackground } = props.snacksTheme.colors
+  let actionActive
+  try {
+    actionActive = tinycolor(actionHover).darken(5).toHexString()
```
wfro: @dcocchia Giving a small lib a try and it feels much better, it was a bigger investment than I expected to try and put together a reliable `darker` function. This isn't exact as far as the color math of the design system, so I can double check with Jordan and see what we can do. If we want to take the same color steps (eg. from green 400 => 500 => 600) for resting => hover => action, I'll need to be more precise and deal with hsl colors.

Another issue - the default snacks theme has the wrong hover color, it's green 700 when it should be green 600 (with the active state being green 700). Can figure that out tomorrow as well.

dcocchia: Interesting that it's wrapped in a try/catch. Is there a high chance of this failing?

wfro:
@dcocchia re: try/catch I was going to ask about that - I feel like it's overkill at this point. I originally added it when i was rolling my own darken function, but might not be necessary now. I'm always on the fence about the right amount of "safeness" with common components like this, any thoughts?

dcocchia:
Given the likelihood that this component will have hundreds of instances on the screen at a time, and that the get styles function may be called frequently, I think we should opt for strong tests and weak, but faster, runtime defensiveness.

wfro:
@dcocchia All of my changes are in, ready to light this candle?

PR #53, opened 2017-09-29
Add a Link component

wfro:
Adds a themer-aware `<Link />` component for hyperlinks.

### DIFF
--- a/src/components/Link/__tests__/Link.spec.js
+++ b/src/components/Link/__tests__/Link.spec.js
@@ -3,7 +3,9 @@ import React from 'react'
 import renderer from 'react-test-renderer'
 import { StyleRoot } from 'radium'
 import { mount } from 'enzyme'
-import { spy, match } from 'sinon'
+import { spy } from 'sinon'
+import themer from '../../../styles/themer'
+import { defaultTheme } from '../../../styles/themer/utils'
 
 describe('Link', () => {
   it('renders without throwing', () => {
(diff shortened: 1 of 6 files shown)


src/components/Link/docs/Link.md
```diff
@@ -0,0 +1,21 @@
+Links!
```
dcocchia: Might want a tad more about why one may need/want a Link component :D

wfro: LOL yes, forgot to fill in this section.

wfro:
@dcocchia pushed another commit with a better docs description than "LInks!" which nullified the last review. I think this is ready to go though. We can possibly tweak the API based on how it gets used.

src/components/Link/__tests__/Link.spec.js
```diff
@@ -0,0 +1,61 @@
+import Link from '../Link'
+import React from 'react'
+import renderer from 'react-test-renderer'
+import { StyleRoot } from 'radium'
+import { mount } from 'enzyme'
+import { spy, match } from 'sinon'
+
+fdescribe('Link', () => {
```
dcocchia: This looks like a typo, should this be `describe`?

src/components/Link/docs/Link.md
```diff
@@ -0,0 +1,22 @@
+The `Link` component is valuable mainly because it's hooked up to Themer.
```
dcocchia: if this is the main value prop of this component, we should add a test or two to test that themer changes change the component output

src/components/Link/__tests__/Link.spec.js
wfro: Ah yeah, this is a test I forgot to un-focus.

PR #54, opened 2017-10-06
Add event hooks into ScrollTrack

dcocchia:
This adds `onBeforeNext`, `onBeforeBack`, `onAfterNext`, `onAfterBack` callbacks to the `ScrollTrack` component. These are useful for things like adding async pagination to a carousel.

### DIFF
--- a/src/components/ScrollTrack/ScrollTrack.js
+++ b/src/components/ScrollTrack/ScrollTrack.js
@@ -18,6 +18,23 @@ class ScrollTrack extends Component {
     /** Manually control left positioning of ScrollTrack */
     leftOverride: PropTypes.number,
 
+    /**
+    * A callback called before sliding to next set.
+    * ** Passed function must return a promsie **
+    * -- will wait for promise resolution before continuing slide.
+    * Use for high levels of control
+    */
+    onBeforeNext: PropTypes.func,
+
+    /**  function to be called before sliding to previous set. */
+    onBeforeBack: PropTypes.func,
+
+    /**  function to be called after sliding to next set. */
+    onAfterNext: PropTypes.func,
+
+    /**  function to be called after sliding to previous set. */
+    onAfterBack: PropTypes.func,
+
     /** Style top level element */
     style: PropTypes.object,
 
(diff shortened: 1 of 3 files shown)


src/components/ScrollTrack/ScrollTrack.js
```diff
@@ -142,21 +159,66 @@ class ScrollTrack extends Component {
...
-    this.setState({ left: nextBack }, this.computeSlideAttributes)
+    const callbackProps = {
+      atStart: nextBack === 0,
+      atEnd: false,
+      slideTo: nextBack,
+      parentWidth,
+      trackWidth
+    }
+
+    onBeforeBack && onBeforeBack(callbackProps)
```
wfro: Total personal preference but I've been liking adding a no-op as a default prop to handle these cases, e.g.

```
ScrollTrack.defaultProps = {
  onBeforeBack: () => {} 
}
```

If we bring in lodash sometime can also do `_.noop`

src/components/ScrollTrack/ScrollTrack.js
```diff
@@ -142,21 +159,66 @@ class ScrollTrack extends Component {
...
+    }
+
+    if (!onBeforeNext) {
+      this.updateLeftValue({
+        left: nextForward,
+        callback: onAfterNext,
+        callbackProps
+      })
+    } else {
+      onBeforeNext(callbackProps).then(() => {
```
wfro: Seems really clean to me. Like we talked about yesterday for a bit, the only other ways i can see to do this is either 1) let rendering components handle pagination 2) have ScrollTrack pass a callback back to it's parent 3) expose a ref on ScrollTrack and have the parent call a method directly to move the scroll position. None of which sound all that exciting :)

src/components/ScrollTrack/docs/ScrollTrack.md
```diff
@@ -129,3 +129,30 @@ Example use case: *A container of item cards that all have a width of 90px and m
...
+      return new Promise((resolve, reject) => {
+          setTimeout(() => {
+            console.log('before next!', props)
+            resolve()
+          }, 1000)
+      })
+    }}
+    onAfterNext={(props) => { console.log('after next!', props) }}
+    onBeforeBack={(props) => { console.log('before back!', props) }}
+    onAfterBack={(props) => { console.log('after back!', props) }}
```
wfro: Not sure what use cases these sync callbacks would have, might be good to add an example or call them out specifically.

dcocchia: A lot of carousel libraries include them so that carousel actions can trigger or control other aspects of a site/component.

src/components/ScrollTrack/ScrollTrack.js
dcocchia: Yeah, I like this. I'll add that in too

PR #56, opened 2017-10-06
Scrolltrack scroll speed and animation customization

dcocchia:


src/components/ScrollTrack/__tests__/__snapshots__/ScrollTrack.spec.js.snap
```diff
@@ -70,6 +70,76 @@ exports[`renders ScrollTrack correctly 1`] = `
...
+      }
+    >
+      <div
+        className="rmq-16b35912"
+        data-radium={true}
+        style={
+          Object {
+            "left": 0,
+            "position": "relative",
+            "transition": "left 1000ms ease-in",
```
wfro: still can't get over how nice this is

wfro:
LGTM, played around with the new props a bit in the styleguide and all seems good.

Making a note to self: the scroll track won't appear on bigger screens because the viewport is too wide 😆 , so we can add something like `maxWidth: '1000px', margin: '0 auto'` on the wrapping div in the example eventually.

PR #59, opened 2017-10-10
Add prop to pass in arbitrary html attributes to NavigationPill(s)

dcocchia:
```
<NavigationPill 
  elementAttributes={{ role: 'tab', 'aria-label: 'this is an aria label' }} 
/>
```


dcocchia:
Should help with https://github.com/instacart/Snacks/pull/45

PR #65, opened 2017-10-11
Add ability to pass custom buttons to ScrollTrack

dcocchia:


dcocchia:
@Eldwick @wfro Would be curious to see if you folks have opinions on this one. Need a way to use elements other than the CircleButton for the scroll track, so came up with this solution. Can you see any issues with it or know of a better way do pull this off?

wfro:
Seems reasonable, if the goal is to take a component instance as a prop and inject some extra props, `cloneElement` is one popular way to do so.

The only two alternatives I've seen are:
* Using children instead of an explicit prop -- I've actually been liking this less and less, especially if you have more than one child element, there's a big implicit dependency on the ordering, e.g.

```
const leftButton = this.props.children[0]
const rightButton = this.props.children[1]
```
* Use a render callback -- this I've been more of a fan of. It returns a little bit more control to the caller, which is good because it provides more flexibility (e.g. where to attach the click listener, etc.), but not so good because it requires more work on part of the caller.

So the code inside `ScrollTrack` might become
```
if (nextButtonElement) {
  return this.renderNextButton({ onClick: this.slideForward, ariaLabel: 'next', style: { ... } })
}
```

And rendering `ScrollTrack` changes to
```
<ScrollTrack
  styles={styles}
  backButtonElement={({ onClick, ariaLabel, style }) => (
    <button onClick={onClick} style={style} aria-label={ariaLabel}>
      <span>back</span>
    </button>
  )}
  {/* Just spreading props is also an option (not accounting for aria-label casing but easy enough to take care of) */}
  nextButtonElement={(props) => (
    <button {...props}>
      <span>next</span>
    </button>
  )}
>
  <p>one</p>
  <p>two</p>
</ScrollTrack>
```

Eldwick:
One concern I have is that unless the `backButtonElement` is a Snacks component or native html tag, the dev implementing it will have to know to implement the hidden passed down props. Imagine a dev wants to create their own custom buttons: 

```
const MyButton = ({ children }) => <button> {children} </button>

<ScrollTrack
  backButtonElement={(
    <MyButton>
      <Icon name="arrowLeft" />
    </MyButton>
  )}
  nextButtonElement={(
    <MyButton>
      <Icon name="arrowRight" />
    </MyButton>
  )}
>
  <div>
    <Icon name="cart" style={styles} />
  </div>
</ScrollTrack>
```

Remembering the `onClick` might be intuitive, but they could easily miss the `aria-label` or `style` prop.

Do we need more flexibility for those buttons other than just styles?

Eldwick:
Maybe something like: 
```
if (nextButtonContent) {
  return (
    <button
      onClick={this.slideForward}
      'aria-label'='next'
      style={...{
        ...slideButtonStyles.default,
        ...slideButtonStyles.right,
        ...{ display: showRightArrow ? 'block' : 'none'},
        ...RightArrow
      }}
    >
      {nextButtonContent}
    </button>
  )
}
```

dcocchia:
Yeah, that's a good solution. I think I'll switch it to that style.

dcocchia:
@Eldwick @wfro Got this using the simpler content style. Ready for approval you think?

src/components/ScrollTrack/ScrollTrack.js
```diff
@@ -17,9 +17,21 @@ class ScrollTrack extends Component {
   static ScrollTrackPropTypes = ScrollTrackPropTypes
 
   static propTypes = {
+    /** Prop for passing in custom button content for back button */
+    backButtonContent: React.PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.Symbol,
```
wfro: Nitpick - does this work (PropTypes.Symbol)? https://reactjs.org/docs/typechecking-with-proptypes.html suggest it's `PropTypes.symbol`, though I've usually used `PropTypes.node` in the past for react elements. I'm actually not sure what the correct way is.

wfro: Nitpick - does this work (PropTypes.Symbol)? https://reactjs.org/docs/typechecking-with-proptypes.html suggest it's `PropTypes.symbol`, though I've usually used `PropTypes.node` in the past for react elements. I'm actually not sure what the correct way is.

wfro:
@dcocchia lgtm, seems like this is the easiest way to get some custom button content without having to replace the entire CircleButton. Easy enough to add animations/icons this way.

src/components/ScrollTrack/ScrollTrack.js
dcocchia: ah yeah, `node` looks much more flexible. Will change to that.

PR #67, opened 2017-10-17
Add TextField to Snacks

nbwar:
Adds the text field component to snacks library

### DIFF
--- a/src/components/Forms/TextField.js
+++ b/src/components/Forms/TextField.js
@@ -1,0 +1,268 @@
+import React             from 'react'
+import PropTypes         from 'prop-types'
+import Radium            from 'radium'
+import { FormComponent } from 'ic-formable'
+import { colors }        from '../../styles'
+import ValidationError   from './ValidationError'
+import FloatingLabel     from './FloatingLabel'
+import TextFieldHint     from './TextFieldHint'
+import ServerError       from './ServerError'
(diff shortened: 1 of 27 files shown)


src/components/Forms/TextField.js
```diff
@@ -0,0 +1,268 @@
...
+
+  state = {
+    hasValue: this.props.defaultValue !== null || Boolean(this.props.value)
+  }
+
+  componentWillMount() {
+    const { name } = this.props
+
+    const uniqueId = `${name}-${Math.floor(Math.random() * 0xFFFF)}`
+    this.uniqueId = uniqueId.replace(/[^A-Za-z0-9-]/gi, '')
```
dcocchia: let's add a comment here for why we're generating this unique id. It also looks possible that the implementer can pass in their own id as a prop, which is what gets used first. Maybe we should make this a default prop rather than always run on mount? That way the docs would also mark it as optional with one generated for you.

nbwar: Only issue with making it a default prop is that we don't have access to `this.props.name` in defaultProps

dcocchia: Yeah, that's true. In that case maybe just a quick check for it in the prop during mount

```js
this.uniqueId = this.props.id || <generate one>
```

PR #68, opened 2017-10-20
Allow equal width HOC accept a function

Eldwick:
Allow equal width HOC accept a function so that `childWidth` can change dynamically.


tjian:
that was super fast. Thanks @Eldwick

dcocchia:
@Eldwick this looks good. Tests are failing though :D

PR #70, opened 2017-10-24
Generic Radio button and Checkbox

julianhr:


dcocchia:
@julianhr is this read to be reviewed?

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+class RadioCheckbox extends Component {
+  static propTypes = {
+    aria          : PropTypes.shape({
+      label         :PropTypes.string,
+    }),
+    btnType       : PropTypes.oneOf(['radio', 'checkbox']),
+    id            : PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
```
dcocchia: We should auto generate an id if one is not provided. We do this in the in the `TextField` component, for example: https://github.com/instacart/Snacks/blob/master/src/components/Forms/TextField.js#L127

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
+import React, { Component, PropTypes }        from 'react'
+import Radium                                 from 'radium'
+import svgRadioBtnBase                        from '../../assets/radioBtnBase.svg'
+import svgRadioBtnChecked                     from '../../assets/radioBtnChecked.svg'
+import svgCheckboxBase                        from '../../assets/checkboxBase.svg'
+import svgCheckboxChecked                     from '../../assets/checkboxChecked.svg'
```
dcocchia: nitpicking: spacing is a bit off

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+      this.setState({styles: this.getStyles()})
+    }
+
+    if (this.props.isChecked !== prevProps.isChecked) {
+      this.setState({isChecked: this.props.isChecked})
+    }
+  }
+
+  onMount() {
+    this.setState({styles: this.getStyles()})
```
dcocchia: this method doesn't seem complex enough to be abstracted out from `componentDidMount`.

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+    this.onMount()
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.isChecked !== prevState.isChecked) {
+      this.setState({styles: this.getStyles()})
+    }
+
+    if (this.props.isChecked !== prevProps.isChecked) {
+      this.setState({isChecked: this.props.isChecked})
```
dcocchia: these `setState`s should probably go in `componentWillReceiveProps` so as not to cause an unnecessary re-render: https://reactjs.org/docs/react-component.html#componentwillreceiveprops

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+    if (btnType === 'checkbox') {
+      if (isChecked) { isCheckedStyle = STYLES.checkboxChecked }
+      finalStyles.button = {...STYLES.checkboxBase, ...isCheckedStyle, ...styles.button}
+    } else {
+      if (isChecked) { isCheckedStyle = STYLES.radioChecked }
+      finalStyles.button = {...STYLES.radioBase, ...isCheckedStyle, ...styles.button}
+    }
+
+    return finalStyles
+  }
```
dcocchia: this method can be a bit more terse:

```js
getStyles() {
  const { btnType, styles } = this.props
  const { isChecked } = this.state
  const isCheckbox = btnType === 'checkbox'
  const isCheckedStyles = isCheckbox ? STYLES.checkboxChecked : STYLES.radioChecked
  const checkedStyles = isChecked ? isCheckedStyles : {}

  return { button: { ...STYLES.checkboxBase, ...checkedStyles, ...styles.button } }
}
```

src/components/Buttons/docs/RadioCheckbox.md
```diff
@@ -0,0 +1,53 @@
+This component outputs a radio button or a checkbox.
+
+The programming logic behind producing these buttons is the same. The difference is in the props `btnType`. It defaults to radio if undefined.
+
+You can define whether the button starts in the pressed on unpressed state through the `isChecked` props.
```
dcocchia: Small typo - `prop` instead of `props`. We should also mention that is can also be controlled past first mount via that prop.

src/components/Buttons/docs/RadioCheckbox.md
```diff
@@ -0,0 +1,53 @@
...
+    <label htmlFor="checkbox1">Checkbox with callback</label>
+  </div>
+  <div style={{display: 'flex'}}>
+    <RadioCheckbox id="checkbox2" btnType='checkbox' isChecked />
+    <label htmlFor="checkbox2">Click me</label>
+  </div>
+</div>
+```
+
+
```
dcocchia: Let's add examples of a controlled radio and checkboxes as well. IOW, controlling the checked state through state within the parent of the radio/checkbox and passing it down as a prop.

src/index.js
```diff
@@ -1,47 +1,54 @@
...
+import ScrollTrack      from './components/ScrollTrack/ScrollTrack'
+import SetStyles        from './styles/SetStyles'
+import spacing          from './styles/spacing'
+import TextField        from './components/Forms/TextField'
+import ServerError      from './components/Forms/ServerError'
+import ValidationError  from './components/Forms/ValidationError'
+import FloatingLabel    from './components/Forms/FloatingLabel'
+import zIndex           from './styles/zIndex'
+import themer           from './styles/themer/index'
+import withTheme        from './styles/themer/withTheme'
```
wfro: I'd prefer we don't align imports to be consistent with the rest of the lib, but open to other arguments

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+    styles      : {}
+  }
+
+  componentDidMount() {
+    this.onMount()
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.isChecked !== prevState.isChecked) {
+      this.setState({styles: this.getStyles()})
```
wfro: It looks like styles do change over time, but they can be computed based off of existing props/state. So better to get styles in `render`, as opposed to storing them on `this.state`.

```
render() {
  const styles = this.getStyles()
  ...
}
```

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+    aria        : {},
+  }
+
+  state = {
+    isChecked   : this.props.isChecked,
+    styles      : {}
+  }
+
+  componentDidMount() {
+    this.onMount()
```
wfro: Per my suggestion somewhere down below of removing styles from `this.state`, we can also drop `this.onMount` entirely

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+
+    return finalStyles
+  }
+
+  onClick = (event) => {
+    const { onClick } = this.props
+    const { isChecked } = this.state
+
+    this.setState({isChecked: !isChecked})
+    if (onClick) { onClick(event, !isChecked) }
```
wfro: I think by convention we've been passing the entire `props` object to click handlers like this (similar to semantic ui)

src/components/Buttons/docs/RadioCheckbox.md
```diff
@@ -0,0 +1,53 @@
...
+```js
+function onClick(event, isChecked) {
+  alert(`This button is ${isChecked ? '' : 'not '}checked`)
+}
+
+<div style={{display: 'flex', flexDirection: 'column'}}>
+  <div style={{display: 'flex', marginBottom: 20}}>
+    <RadioCheckbox id="radio1" onClick={onClick} />
+    <label htmlFor="radio1">Click me with callback</label>
```
wfro: Would it make sense to allow users to pass a label directly in to the component? It seems like a common use case. Could be either a string or a rendered element, e.g. `<Checkbox label={<label>Make my profile visible</label>} />` (https://react.semantic-ui.com/modules/checkbox#checkbox-example-shorthand-element)

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+  checkboxChecked: {
+    backgroundImage: `url(${svgCheckboxChecked.url})`,
+    ':active': {
+      backgroundImage: `url(${svgCheckboxChecked.url})`,
+      border: '0 solid transparent'
+    }
+  }
+}
+
+class RadioCheckbox extends Component {
```
wfro: I'd be a fan of exporting multiple components, e.g. `Checkbox` and `Radio`, with `Radio` being a small wrapper around `Checkbox`. I like it mostly because the API feels more self-explanatory. Docs could be short for `Radio` as well, just a small example and maybe link back to the checkbox docs

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
+import React, { Component, PropTypes }        from 'react'
```
wfro: `Checkbox` might make sense as it's on top level components folder, i.e. `src/components/Checkbox`

src/components/Buttons/RadioCheckbox.js
wfro: One option for avoiding the `if (onClick)` check is to register a noop as a default props, e.g.
```
const noop = () => {}

static defaultProps = {
  onClick: noop
}
```

src/components/Buttons/__tests__/RadioCheckbox.spec.js
```diff
@@ -0,0 +1,51 @@
+import RadioCheckbox from '../RadioCheckbox'
+import React from 'react'
+import renderer from 'react-test-renderer'
+import { mount } from 'enzyme'
+import sinon from 'sinon'
```
wfro: Unused sinon/enzyme imports

wfro: Although adding a test for the `onClick` handler being called may be good

src/components/Buttons/docs/RadioCheckbox.md
wfro: The markup for the component would have to change slightly to something like this (pardon awkward spacing)

```
<div
        onChange={this.handleClick}
        onClick={this.handleClick}
      >
        <input
          checked={checked}
          name={name}
          type={type}
          value={value}
        />
        {*/ some kind of label (optional) */}
      </div>
```

package.json
```diff
@@ -69,7 +70,9 @@
   },
   "jest": {
     "moduleNameMapper": {
-      "^utils": "<rootDir>/src/utils"
+      "^utils": "<rootDir>/src/utils",
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/scripts/test/fileMock.js",
+      "\\.(css|less)$": "<rootDir>/scripts/test/styleMock.js"
```
wfro: What's this one for out of curiousity? I think we're still css-free for snacks so far

src/components/Buttons/RadioCheckbox.js
dcocchia: Agreed - that seems simpler and more obvious for users of the library

package.json
dcocchia: I added this for the svgs mostly. This is the generic style mocking that jest suggests

src/components/Buttons/RadioCheckbox.js
julianhr: This reason for this comes from the linter in the `carrot` repo:

screen shot 2017-10-25 at 5 01 33 pm

The way I interpret this rule is to avoid firing various `setState` calls in one method. I got used to it and I always now take this approach. However, now that you bring it up I noticed that this rule is not turned on in Snacks.

What do you think, should I not follow that rule here?

This comment also addresses https://github.com/instacart/Snacks/pull/70/files#r147008676 @wfro

src/components/Buttons/RadioCheckbox.js
julianhr: Also, radio buttons, checkboxes and labels were one of the major problems found in our accessibility audit. That is the principal reason for this check.

src/components/Buttons/RadioCheckbox.js
dcocchia: Ah, right. This is to avoid setting state immediately on mount, which causes an immediate re-render. This should instead go in componentWillMount. This way only one render is occurs on mount.

src/components/Buttons/RadioCheckbox.js
dcocchia: I would say we should stop requiring it, then.

src/components/Buttons/RadioCheckbox.js
dcocchia: Yep. But I think within snacks we don't do this sort of spacing at all

src/components/Buttons/docs/RadioCheckbox.md
dcocchia: It should already accept es6. We're using it in a few examples already - for example: https://instacart.github.io/Snacks/#icons

src/components/Buttons/RadioCheckbox.js
dcocchia: It may feel like more work for us, but it will feel like a more clear and easier to use library from the outside. That's what we're aiming for :)

src/components/Buttons/docs/RadioCheckbox.md
wfro: 👍  taking a look now. this was mostly motivated by my assumption that most users of `RadioCheckbox` would have to create their own `label` element, which if true, means it makes sense for us to do that piece for them in snacks. I think it also lets us ensure labels always get applied because we're doing it at the snacks level. For the most part ripping directly from https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Checkbox/Checkbox.js#L222

src/components/Buttons/RadioCheckbox.js
wfro: Any thoughts from https://github.com/instacart/Snacks/pull/70/files#r147008560? Unless I'm missing something (very possible :))I think we can drop calls to `this.setState({ styles: this.getStyles() })` entirely

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+
+  static defaultProps = {
+    isChecked   : false,
+    btnType     : 'radio',
+    styles      : {},
+    aria        : {},
+  }
+
+  state = {
+    isChecked   : this.props.isChecked,
```
Eldwick: We should consider adding an auto controlled abstraction, in slightly more complicated components, allowing state to be controlled here or in the parent can start to get really confusing:

https://github.com/Semantic-Org/Semantic-UI-React/blob/160223f82c7d2e0e152d3fd1f176e315f27b3a66/src/lib/AutoControlledComponent.js

src/components/Buttons/RadioCheckbox.js
Eldwick: Yeah, I don't think there is any reason to store styles in state

src/components/Buttons/RadioCheckbox.js
Eldwick: I would add in the checked state:

```
onClick(event, { ...this.props, isChecked: !isChecked })
```

src/components/Buttons/RadioCheckbox.js
julianhr: Yeah, I think you're right. Updating.

src/components/Buttons/RadioCheckbox.js
julianhr: ok, updating.

src/components/Buttons/RadioCheckbox.js
```diff
@@ -0,0 +1,147 @@
...
+
+const BASE_STYLES = {
+  input: {
+    margin: '0 10px -7px 0',
+    backgroundSize: 'contain',
+    backgroundRepeat: 'no-repeat',
+    backgroundPosition: 'top left',
+    backgroundColor: 'transparent',
+    appearance: 'none',
+    '-moz-appearance': 'none',
```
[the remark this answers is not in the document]
julianhr: I don't think we have an option here unless we include a prefixer library. Without those prefixes that code will not work in all browsers. cc @dcocchia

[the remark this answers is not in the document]
julianhr: I don't think we have an option here unless we include a prefixer library. I did bring this up with Dominic but he mentioned I would have to prefix it. Without those prefixes this code will not work in all browsers. cc @dcocchia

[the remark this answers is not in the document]
dcocchia: Radium claims to auto-prefix: https://formidable.com/open-source/radium/ so long as a component is wrapped in the radium HOC

src/components/Buttons/RadioCheckbox.js
dcocchia: I'd be into this if we have a strong enough use case. If we run into another issue like this, let's do it then 👍

src/components/Buttons/RadioCheckbox.js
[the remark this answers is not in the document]
julianhr: The component is wrapped in the radium HOC but I don't think the auto-prefixing works. I tried it and it stopped working:

With prefixes: screen shot 2017-10-25 at 10 58 31 pm


Without prefixes: screen shot 2017-10-25 at 10 58 50 pm

julianhr:
Mejor modifications. Separate Radio and Checkbox buttons. Please help review again.

src/base/RadioCheckboxBase.js
```diff
@@ -0,0 +1,137 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    const { isChecked } = nextProps
+    if (isChecked !== this.props.isChecked) { this.setState({isChecked}) }
+  }
+
+  getStyles() {
+    const { styles, assets } = this.props
+    const { isChecked } = this.state
+    const { btnBkg } = assets
```
Eldwick: Is assets always going to be defined as an object? If so, we need to make it a required prop or give it a default prop value

julianhr: Good point. Updating.

dcocchia:
This looks good to me. @Eldwick any issues remaining for you?

julianhr:
@Eldwick this branch is already out of sync with master. It'd be great if you can provide feedback so I can merge this as soon as possible.

dcocchia:
Going to merge this one. We can always follow up with polish if we need to :)

wfro:
:tada::tada::tada::tada::tada::tada::tada:

julianhr:
👌

PR #72, opened 2017-10-30
moving icformable into snacks

nbwar:
Moving `Form` and `FormComponent` from `ICFormable` into snacks to make it easier for other people to add/edit functionality in the future as well as consolidate component libraries & dependencies.


src/components/Forms/docs/FormComponent.md
```diff
@@ -0,0 +1,24 @@
+Read more @ https://github.com/nbwar/icformable
```
dcocchia: We should probably move those docs over as well, at least in part

src/components/Forms/Form.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  static childContextTypes = {
+    ICFormable: PropTypes.object
+  }
+
+  constructor() {
+    super()
+    this.model          = {}
+    this.formComponents = {}
```
dcocchia: since we define `state` directly on the class as an instance variable, we should do `model` and `formComponents ` that way too I think

src/components/Forms/docs/FormComponent.md
nbwar: `FormComponent` didn't show up in the docs menu. Not sure if it's because it is an HOC.

dcocchia: Oh I mean rather than sending people off to another url for docs, we should have the docs here I think

PR #85, opened 2017-11-17
Radio groups now possible

julianhr:
Now it's possible to create radio groups where only one button can be selected at a time.
Updated the Radio and Checkbox components in order to support radio button groups.


dcocchia:
Seeing some style property errors:
screen shot 2017-11-16 at 5 12 32 pm

dcocchia:
The `RadioGroup` should have an `onChange` callback, so that users of the group can get the selected value. It's helpful to imagine these being used in a form with many inputs, for example.

It would be good try and create a form using `ICFormable` with some inputs and a radio group that alerts all the values of the entire form on submit both so that users have the example available, but also so we can ensure that the use case works well.

julianhr:
@dcocchia Unfortunately I don't think we can do much regarding these prefixed `appearance` attributes. We need them so each browser styles the radio button correctly:

screen shot 2017-11-16 at 5 46 49 pm

julianhr:
@dcocchia good idea on the `onChange` callback. Implementing.

dcocchia:
We can do browser prefixing, but it needs to be done the js styles way. `webkitAppearance`, `mzAttribute`, etc

julianhr:
@dcocchia requested changes made. Please help review again.

PR #89, opened 2017-11-28
[buttons] Add text-align center to base Button

wfro:
Should probably have included this from the get-go, but better late than never :)


src/components/Buttons/Button.js
```diff
@@ -10,6 +10,7 @@ import { darken } from '../../utils'
 const noop = () => {} // eslint-disable-line no-empty-function
 
 const baseStyles = {
+  textAlign: 'center',
```
julianhr: Since you are using `flex`, I suggest turning this to `justifyContent: 'center'`:
https://css-tricks.com/snippets/css/a-guide-to-flexbox/#article-header-id-6
The reason is that using `text-align` may not work properly when nesting other things, like images or Icons inside the button and `justifyContent` works correctly across all types of nested elements.

wfro: @julianhr great call, thanks. I don't think text-align: center even does what I want here.

wfro: On second thought, inline-flex is what's causing the issue in the first place, so I'm gonna close this for now and see if I can solve it there.

PR #91, opened 2017-12-01
Menu, DropdownMenu, Select

nbwar:
Adding Menu, DropdownMenu, and Select FormComponent.

Menu:
screen shot 2017-12-01 at 3 21 59 pm

DropdownMenu
shb4yassvj

Select
jsavxnos5w

### DIFF
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -216,16 +216,16 @@ class Select extends React.PureComponent {
     })
   }
 
-  handleSelect = (option) => {
+  handleSelect = (e, option) => {
     const { onSelect, selectedOption } = this.props
 
     // For manual control
     if (!selectedOption) {
       this.setState({selectedOption: option}, () => {
-        onSelect(option)
+        onSelect(e, option)
       })
     } else {
-      onSelect(option)
+      onSelect(e, option)
     }
   }
 
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -105,7 +105,7 @@ const styles = {
 
 @FormComponent
 @Radium
-class Select extends React.Component {
+class Select extends React.PureComponent {
   static propTypes = {
     /** Name of the field */
     name               : PropTypes.string.isRequired,
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -104,6 +104,8 @@ const styles = {
   }
 }
 
+const noOp = () => {} // eslint-disable-line no-empty-function
+
 @FormComponent
 @Radium
 class Select extends React.PureComponent {
@@ -170,11 +172,11 @@ class Select extends React.PureComponent {
   static defaultProps = {
     disabled    : false,
     defaultOption: null,
-    onFocus: () => {}, // eslint-disable-line no-empty-function
-    onBlur: () => {}, // eslint-disable-line no-empty-function
-    onOpen: () => {}, // eslint-disable-line no-empty-function
-    onClose: () => {}, // eslint-disable-line no-empty-function
-    onSelect: () => {} // eslint-disable-line no-empty-function
+    onFocus: noOp, // eslint-disable-line no-empty-function
+    onBlur: noOp, // eslint-disable-line no-empty-function
+    onOpen: noOp, // eslint-disable-line no-empty-function
+    onClose: noOp, // eslint-disable-line no-empty-function
+    onSelect: noOp // eslint-disable-line no-empty-function
   }
 
   state = {
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -14,7 +14,6 @@ import Icon            from '../Icon/Icon'
 
 const styles = {
   root: {
-    cursor: 'auto',
     display: 'inline-block',
     position: 'relative',
     width: '343px',
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -192,30 +192,30 @@ class Select extends React.Component {
     return (this.state.selectedOption && this.state.selectedOption.value) || null
   }
 
-  _handleClose = () => {
+  handleClose = () => {
     this.trigger.focus()
     this.props.onClose()
   }
 
-  _handleRequestChange = (open) => {
+  handleRequestChange = (open) => {
     if (!this.props.disabled) {
       this.setState({isOpen: open})
     }
   }
 
-  _handleFocus = (e) => {
+  handleFocus = (e) => {
     this.setState({isFocused: true}, () => {
       this.props.onFocus(e)
     })
   }
 
-  _handleBlur = (e) => {
+  handleBlur = (e) => {
     this.setState({isFocused: false}, () => {
       this.props.onBlur(e)
     })
   }
 
-  _handleSelect = (option) => {
+  handleSelect = (option) => {
     const { onSelect, selectedOption } = this.props
 
     // For manual control
@@ -243,8 +243,8 @@ class Select extends React.Component {
 
     return (
       <div
-        onFocus={this._handleFocus}
-        onBlur={this._handleBlur}
+        onFocus={this.handleFocus}
+        onBlur={this.handleBlur}
       >
         <div
           style={[
@@ -331,10 +331,10 @@ class Select extends React.Component {
           <DropdownMenu
             triggerElement={this.renderTrigger()}
             open={isOpen}
-            onRequestChange={this._handleRequestChange}
-            onSelect={this._handleSelect}
+            onRequestChange={this.handleRequestChange}
+            onSelect={this.handleSelect}
             menuProps={{style: styles.menu}}
-            onClose={this._handleClose}
+            onClose={this.handleClose}
             onOpen={onOpen}
           >
             {children}
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -32,7 +32,7 @@ const styles = {
     fontSize: '16px',
     height: '56px',
     margin: '0',
-    padding: '25px 8px 8px 8px',
+    padding: '25px 25px 8px 8px',
     outline: 'none',
     position: 'relative',
     WebkitOpacity: 1,
@@ -56,11 +56,12 @@ const styles = {
     width: '100%'
   },
   iconContainer: {
-    paddingLeft: '1px',
     display: 'flex',
     flexDirection: 'column',
     userSelect: 'none',
-    float: 'right'
+    position: 'absolute',
+    right: '8px',
+    top: '22px'
   },
   icon: {
     color: colors.GREEN_500,
--- a/src/components/Forms/Select.js
+++ b/src/components/Forms/Select.js
@@ -1,0 +1,351 @@
+import React           from 'react'
+import PropTypes       from 'prop-types'
+import Radium          from 'radium'
+import { colors }      from '../../styles'
+import FormComponent   from '../Forms/FormComponent'
+import DropdownMenu    from '../Menus/DropdownMenu'
+import ValidationError from '../Forms/ValidationError'
+import ServerError     from '../Forms/ServerError'
+import TextFieldHint   from '../Forms/TextFieldHint'
+import FloatingLabel   from '../Forms/FloatingLabel'
+import Icon            from '../Icon/Icon'
+
+/* eslint jsx-a11y/no-noninteractive-tabindex: 0 */
+
+const styles = {
+  root: {
+    cursor: 'auto',
+    display: 'inline-block',
+    position: 'relative',
+    width: '343px',
+  },
+  triggerContainer: {
+    borderRadius: '4px',
+    position: 'relative'
+  },
+  trigger: {
+    backgroundColor: '#FFF',
+    border: `solid 1px ${colors.GRAY_74}`,
+    borderRadius: '4px',
+    boxSizing: 'border-box',
+    color: colors.GRAY_20,
+    fontSize: '16px',
+    height: '56px',
+    margin: '0',
+    padding: '25px 8px 8px 8px',
+    outline: 'none',
+    position: 'relative',
+    WebkitOpacity: 1,
+    WebkitTapHighlightColor: 'rgba(0,0,0,0)',
+    ':hover': {
+      cursor: 'pointer'
+    }
+  },
+  triggerDisabled: {
+    border: `1px dashed ${colors.GRAY_74}`,
+    backgroundColor: colors.GRAY_93,
+    color: colors.GRAY_46,
+    ':hover': {
+      cursor: 'not-allowed'
+    }
+  },
+  highlight: {
+    border: `1px solid ${colors.GREEN_500}`
+  },
+  menu: {
+    width: '100%'
+  },
+  iconContainer: {
+    paddingLeft: '1px',
+    display: 'flex',
+    flexDirection: 'column',
+    userSelect: 'none',
+    float: 'right'
+  },
+  icon: {
+    color: colors.GREEN_500,
+    transition: 'transform 100ms',
+    userSelect: 'none'
+  },
+  iconError: {
+    color: colors.RED_700
+  },
+  iconOpen: {
+    transform: 'rotate(180deg)'
+  },
+  iconDisabled: {
+    color: colors.GRAY_46
+  },
+  labelContainer: {
+    whiteSpace: 'nowrap',
+    textOverflow: 'ellipsis',
+    overflow: 'hidden',
+    userSelect: 'none',
+    display: 'inline-block'
+  },
+  floatingLabel: {
+    position: 'absolute',
+    top: 0,
+    left: 0
+  },
+  error: {
+    border: `1px solid ${colors.RED_700}`,
+    backgroundColor: '#FDE6EB'
+  },
+  fullWidth: {
+    width: '100%'
+  },
+  halfWidth: {
+    width: '162px'
+  }
+}
+
+@FormComponent
+@Radium
+class Select extends React.Component {
+  static propTypes = {
+    /** Name of the field */
+    name               : PropTypes.string.isRequired,
+    /** Children are passed to Menu and should be MenuItems or MenuDivider */
+    children           : PropTypes.node,
+    /** DefaultOption */
+    defaultOption       : PropTypes.shape({
+      label: PropTypes.string,
+      value: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+        PropTypes.bool
+      ])
+    }),
+    /** Disable the text field */
+    disabled           : PropTypes.bool,
+    /** Text of label that will animate when TextField is focused */
+    floatingLabelText  : PropTypes.string,
+    /** Sets width to 100% */
+    fullWidth          : PropTypes.bool,
+    /** Sets width to 162px */
+    halfWidth          : PropTypes.bool,
+    /** FormComponent error for validation */
+    hasError           : PropTypes.bool,
+    /** Hint text will show up when the Select is open and there is no value */
+    hintText           : PropTypes.string,
+    /** Uniq id for Select */
+    id                 : PropTypes.string,
+    /** Handled by FormComponent after running built in validations */
+    isValid            : PropTypes.bool,
+    /** onOpen callback */
+    onOpen             : PropTypes.func,
+    /** onClose callback */
+    onClose            : PropTypes.func,
+    /** onFocus callback */
+    onFocus            : PropTypes.func,
+    /** onSelect callback returns option object {label: , value:}*/
+    onSelect           : PropTypes.func,
+    /** onBlur callback */
+    onBlur             : PropTypes.func,
+    /** Mark the field as required.  */
+    required           : PropTypes.bool,
+    /** Control the component selection manually */
+    selectedOption     : PropTypes.shape({
+      label: PropTypes.string,
+      value: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+        PropTypes.bool
+      ])
+    }),
+    /** Error from server to show ServerError message */
+    serverError        : PropTypes.string,
+    /** Wrapper styles, mainly used for custom width */
+    style              : PropTypes.object,
+    /** Text to show for validation error  */
+    validationErrorText: PropTypes.string,
+    /** Value will make TextField a controlled component  */
+  }
+
+  static defaultProps = {
+    disabled    : false,
+    defaultOption: null,
+    onFocus: () => {}, // eslint-disable-line no-empty-function
+    onBlur: () => {}, // eslint-disable-line no-empty-function
+    onOpen: () => {}, // eslint-disable-line no-empty-function
+    onClose: () => {}, // eslint-disable-line no-empty-function
+    onSelect: () => {} // eslint-disable-line no-empty-function
+  }
+
+  state = {
+    isOpen: false,
+    selectedOption: this.props.defaultOption || this.props.selectedOption || null
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { selectedOption } = nextProps
+    if (typeof selectedOption !== 'undefined' && selectedOption !== this.state.selectedOption) {
+      this.setState({selectedOption: selectedOption})
+    }
+  }
+
+  getValue = () => {
+    return (this.state.selectedOption && this.state.selectedOption.value) || null
+  }
+
+  _handleClose = () => {
+    this.trigger.focus()
+    this.props.onClose()
+  }
+
+  _handleRequestChange = (open) => {
+    if (!this.props.disabled) {
+      this.setState({isOpen: open})
+    }
+  }
+
+  _handleFocus = (e) => {
+    this.setState({isFocused: true}, () => {
+      this.props.onFocus(e)
+    })
+  }
+
+  _handleBlur = (e) => {
+    this.setState({isFocused: false}, () => {
+      this.props.onBlur(e)
+    })
+  }
+
+  _handleSelect = (option) => {
+    const { onSelect, selectedOption } = this.props
+
+    // For manual control
+    if (!selectedOption) {
+      this.setState({selectedOption: option}, () => {
+        onSelect(option)
+      })
+    } else {
+      onSelect(option)
+    }
+  }
+
+  renderTrigger() {
+    const {
+      disabled,
+      required,
+      hasError,
+      hintText,
+      floatingLabelText
+    } = this.props
+
+    const { isOpen, selectedOption, isFocused } = this.state
+    const hasValue = selectedOption !== null
+    const showHintText = hintText && !hasValue && isOpen
+
+    return (
+      <div
+        onFocus={this._handleFocus}
+        onBlur={this._handleBlur}
+      >
+        <div
+          style={[
+            styles.trigger,
+            disabled && styles.triggerDisabled,
+            (isOpen || isFocused) && !hasError && styles.highlight,
+            !disabled && hasError && styles.error
+          ]}
+          aria-required={required}
+          aria-invalid={hasError}
+          tabIndex={0}
+          ref={ (node) => this.trigger = node }
+        >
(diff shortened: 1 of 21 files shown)


src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+            show={!disabled && !isValid && !serverError}
+            inputId={id}
+          />
+        </div>
+      </div>
+    )
+  }
+}
+
+export default Select
```
dcocchia: Hovering on the text in the select gives me a cursor rather than a pointer
screen shot 2017-12-04 at 3 30 49 pm

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+    switch(event.key) {
+      case 'ArrowDown':
+        event.preventDefault()
+        this.incrementTabIndex()
+        break
+      case 'ArrowUp':
+        event.preventDefault()
+        this.decrementTabIndex()
+        break
+    }
```
dcocchia: nitpicking, prefer `{}` on cases as well

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+}
+
+class Menu extends React.Component {
+  static propTypes = {
+    /** aria-hidden HTML attribute */
+    ariaHidden: PropTypes.bool,
+    /** MenuItems or Divider */
+    children: PropTypes.node.isRequired,
+    /** The maximum height of the Menu when it is displayed */
+    maxHeight: PropTypes.number,
```
dcocchia: I'm guessing this can't be controlled via the style object. Just curious what stops that from working?

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+
+    return React.Children.map(children, (child) => {
+      if (!React.isValidElement(child)) {
+        return null
+      } else if (child.type && child.type.displayName === 'MenuItem') {
+        const component = React.cloneElement(child, {
+          index: index,
+          focus: currentTabIndex === index,
+          onClick: this._handleClick,
+          onMenuItemFocus: this._handleMenuItemFocus
```
dcocchia: Can we add code comments around how this focus pattern works?

src/components/Menus/MenuItem.js
```diff
@@ -0,0 +1,207 @@
+import React             from 'react'
+import PropTypes         from 'prop-types'
+import Radium            from 'radium'
+import { colors }        from '../../styles'
+import Icon              from '../Icon/Icon'
+
+/* eslint jsx-a11y/no-noninteractive-tabindex: 0 */
```
dcocchia: Does eslint not do well with the default prop thing?

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+    if (typeof selectedOption !== 'undefined' && selectedOption !== this.state.selectedOption) {
+      this.setState({selectedOption: selectedOption})
+    }
+  }
+
+  getValue = () => {
+    return (this.state.selectedOption && this.state.selectedOption.value) || null
+  }
+
+  _handleClose = () => {
```
Eldwick: Why the `_` convention here? I think the prefix `handle` implies enough that this is a function within the component.

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { selectedOption } = nextProps
+    if (typeof selectedOption !== 'undefined' && selectedOption !== this.state.selectedOption) {
+      this.setState({selectedOption: selectedOption})
+    }
+  }
+
+  getValue = () => {
```
Eldwick: There's also the getter syntax for es6 classes:
```
get value () {
  return (this.state.selectedOption && this.state.selectedOption.value) || null
}
```
then you can run the function with just `this.value`

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+      hintText,
+      floatingLabelText
+    } = this.props
+
+    const { isOpen, selectedOption, isFocused } = this.state
+    const hasValue = selectedOption !== null
+    const showHintText = hintText && !hasValue && isOpen
+
+    return (
+      <div
```
Eldwick: Looks like a `style` prop is passed in, should it be applied here?

jordanscode:
How would you all feel of we added a min-width to this? (say 208px?)

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  state = {
+    currentTabIndex: null
+  }
+
+  _handleClick = (option, index) => {
+    const { onSelect } = this.props
+
+    onSelect && onSelect(option, index)
```
Eldwick: I think we are trying to implement a pattern where we have a consistent/predictable callback signature:
```
callback(event, {...props, ...state })
```

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+  _handleSelect = (option) => {
+    const { onSelect, selectedOption } = this.props
+
+    // For manual control
+    if (!selectedOption) {
+      this.setState({selectedOption: option}, () => {
+        onSelect(option)
+      })
+    } else {
+      onSelect(option)
```
Eldwick: I made a comment below, but we discussed trying to standardize callback signatures to take the event and props/state to allow the parent components of `Snack` components to easily predict how callbacks will be used.

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+    }
+  }
+
+  renderTrigger() {
```
Eldwick: This seems like a pretty big component `renderTrigger` - I think it probably warrants a separate component. I think ideally we should be shooting for sub 100 line components if possible.

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+    width: '100%'
+  },
+  halfWidth: {
+    width: '162px'
+  }
+}
+
+@FormComponent
+@Radium
+class Select extends React.Component {
```
Eldwick: I think this can be a `PureComponent`

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+      this.setState({selectedOption: selectedOption})
+    }
+  }
+
+  getValue = () => {
+    return (this.state.selectedOption && this.state.selectedOption.value) || null
+  }
+
+  _handleClose = () => {
+    this.trigger.focus()
```
Eldwick: Why do we focus the input after close?

src/components/Forms/TextField.js
```diff
@@ -254,7 +250,7 @@ class TextField extends React.Component {
             onBlur={this.handleInputBlur}
             onChange={this.handleInputChange}
             onFocus={this.handleInputFocus}
-            onKeyDown={onKeyDown && onKeyDown()}
+            onKeyDown={ (e) => onKeyDown(e) }
```
Eldwick: We should break this out to `handleKeyDown` like the handlers above this line. Anonymous functions generated in the render method will register as a false prop change on render.

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
+import React             from 'react'
+import PropTypes         from 'prop-types'
+import Radium            from 'radium'
+import Menu              from './Menu'
+
+const styles = {
+  menuContainer: {
+    zIndex: 9999,
```
Eldwick: I think we have a zIndex style import within Snacks

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+    /** Callback function fired when a MenuItem is selected */
+    onSelect: PropTypes.func,
+    /** Customize style of menu parent */
+    style: PropTypes.shape({}),
+    /** Props passed to Menu component */
+    menuProps: PropTypes.shape({})
+  }
+
+  static defaultProps = {
+    open: null
```
Eldwick: This doesn't seem necessary, can't we just check `nextProps.open !== undefined` within `componentWillReceiveProps`?

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+        }
+        break
+    }
+  }
+
+  controlledOpen() {
+    return this.props.open !== null
+  }
+
+  open() {
```
Eldwick: Maybe we should stick with the handler convention of prefixing this with `handle`?

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+        this.close()
+      }
+    }
+  }
+
+  _handleSelect = (option) => {
+    const { onSelect, onRequestChange } = this.props
+
+    if (this.controlledOpen()) {
+      onRequestChange && onRequestChange(false)
```
Eldwick: What is the use case for `onRequestChange`?

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+    } else {
+      // if multi then do not close
+      this.close()
+    }
+
+    onSelect && onSelect(option)
+  }
+
+  _handleClick = (event) => {
+    event.preventDefault()
```
Eldwick: Are we sure we always want to prevent default? Seems like we might want to leave that decision to the parent component

dcocchia:
@jordanscode min-width seems ok to me, so long as we maintain a way to override it. ( the style prop )

src/components/Forms/Select.js
dcocchia: Otherwise focus goes to body, which is 🚫  for a11y.

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+    })
+  }
+
+  renderTriggerElement() {
+    const { triggerElement } = this.props
+    const { open } = this.state
+
+    if (triggerElement) {
+      return React.cloneElement(triggerElement, {
+        onClick: this._handleClick,
```
Eldwick: This seems like unexpected behavior. I wouldn't assume that the `triggerElement` I passed in would lose it's click handler when passed into the component, can we just surround the `triggerElement` with a another div that we give the click handler too?

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+      this.open()
+    } else {
+      this.close()
+    }
+    onClick && onClick(event)
+  }
+
+  _handleMenuBlur = () => {
+    const { onRequestChange } = this.props
+    // Timeout ensures click event has correct state when attempting to close
```
Eldwick: What is the event you are worried about here?

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  _handleBlur = (event) => {
+    const { onBlur } = this.props
+    const currentTarget = event.currentTarget
+
+    setTimeout( () => {
```
Eldwick: Why is the timeout necessary and what is the conditional doing?

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+    return menuItemChildren
+  }
+
+  renderChildren() {
+    const { children } = this.props
+    const { currentTabIndex } = this.state
+    let index = 0
+
+    return React.Children.map(children, (child) => {
+      if (!React.isValidElement(child)) {
```
Eldwick: I think we should fail hard here, if it's not a valid element the developer should catch it in development

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+
+    return menuItemChildren
+  }
+
+  renderChildren() {
+    const { children } = this.props
+    const { currentTabIndex } = this.state
+    let index = 0
+
+    return React.Children.map(children, (child) => {
```
Eldwick: Kinda feels like `children` is the wrong way to pass the necessary information into this component. Since the only really "valid" children are `MenuItems` we should pass probably pass down `menuItems` as an array of hashes that `Menu` will consume and render. One of the configs for a given `menuItem` could be `divider: true` which would append a `Divider`.

With this pattern you don't have to do all the gymnastics of cloning the children, checking their validity, or trying to do complicated logic to increment the next selected item.

This way you can also more easily validate the `menuItems` passed down to ensure they are in the shape you expect.

src/components/Menus/MenuDivider.js
```diff
@@ -0,0 +1,26 @@
...
+  margin: '8px 0',
+  height: 1,
+  border: 'none'
+}
+
+const MenuDivider = (props) => {
+  const { style } = props
+
+  return (
+    <hr style={Object.assign(styles, style)}/>
```
Eldwick: This will change the `styles` for all future `MenuDividers` since it will mutate the constant above. I would prefer the syntax `{... styles, ... style}` in the future to avoid this issue.

src/components/Menus/MenuItem.js
```diff
@@ -0,0 +1,207 @@
...
+      disabled,
+      index,
+      onClick,
+      value,
+      label
+    } = this.props
+
+    const option = {value: value, label: label}
+
+    !disabled && onClick && onClick(option, index)
```
Eldwick: Just another reminder to standardize the way we send information to callbacks

src/components/Menus/MenuItem.js
```diff
@@ -0,0 +1,207 @@
...
+
+  _handleFocus = (event) => {
+    const { onFocus, onMenuItemFocus, index } = this.props
+    onMenuItemFocus && onMenuItemFocus(index)
+    onFocus && onFocus(index, event)
+  }
+
+  renderLeftIcon() {
+    const { leftIcon, leftIconStyles } = this.props
+    let iconComponent = null
```
Eldwick: I'm a big fan of early returns to simplify logic/avoid mutating variables:
```
if (!leftIcon) { return }
const iconComponent = React.isValidElement(leftIcon) ? leftIcon : <Icon name={leftIcon} style={[styles.leftIconStyles, leftIconStyles]} />
```

Eldwick:
Also in the future would be awesome to see slightly smaller PRs so we can provide feedback in smaller chunks :)

src/components/Menus/MenuItem.js
Eldwick: Actually I think we should reverse the conditional so that we get better error messages if `leftIcon` is not a valid component:
```
const iconComponent = typeof leftIcon === 'String'
  ? <Icon name={leftIcon} style={[styles.leftIconStyles, leftIconStyles]} />
  : leftIcon
```

src/components/Forms/Select.js
```diff
@@ -0,0 +1,351 @@
...
+    }),
+    /** Disable the text field */
+    disabled           : PropTypes.bool,
+    /** Text of label that will animate when TextField is focused */
+    floatingLabelText  : PropTypes.string,
+    /** Sets width to 100% */
+    fullWidth          : PropTypes.bool,
+    /** Sets width to 162px */
+    halfWidth          : PropTypes.bool,
+    /** FormComponent error for validation */
```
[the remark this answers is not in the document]
nbwar: Neither `fullWidth` or `halfWidth` are the default. fullWidth sets width to 100% vs `halfWidth` is half the default width of 343px

src/components/Menus/DropdownMenu.js
nbwar: `onRequestChange` is to manually control the open/close of the dropdown

src/components/Menus/DropdownMenu.js
nbwar: When a MenuItem is selected the Menu is blur

src/components/Menus/Menu.js
nbwar: This specific callback needs to pass the selected option. Would this just be appended to the end of the arguments? Not sure i like this pattern.

src/components/Menus/DropdownMenu.js
Eldwick: Why isn't that handle via `handleSelect`?

nbwar: `handleSelect` default behavior is to close when a MenuItem is selected. However, if you want to control the open and close manually via props that is also possible. See `Select`.

src/components/Menus/DropdownMenu.js
Eldwick: Sorry, I still don't understand the issue - the comment above says that the timeout ensures the click event will have the correct state. What is the click event and what is the correct state that is not correct without the timeout?

src/components/Menus/Menu.js
nbwar: Even if the children were passed as a props they would still need to be cloned to pass `onClick` listener so the Menu can can respond to the onClick callback and do it's thing. I don't see how this will change whether it is children or props. It essentially doing the same thing just passed to Menu differently.

Eldwick: You wouldn't pass the children as react components, you would pass them as native objects:
```
menuItems = [{label: 'United States', value:'US', divider: true}, {label: 'Canada', value: 'CA' }]
```

I made the arguments above, but basically by passing the data down this way will let you  operate on the data more easily.

src/components/Menus/Menu.js
Eldwick: Well it seems like this is actually the wrong place to put the click handler, `Menu` should just pass down `onSelect` to `MenuItem` and `MenuItem` should pass it's props to the callback. Then you have the label and value within the properties argument:
```
class MenuItem {
  handleClick = (e) => {
    nSelect && onSelect(e, this.props)
   }
}
```

Eldwick: For something like an input element you would extract the value from the event and pass it along with the props:
```
class SomeSnacksCustomInput {
  handleClick = (e) => {
    const { value } = e.target
    onSelect && onSelect(e, {...this.props, value })
   }
}
```

src/components/Menus/Menu.js
nbwar: I thought of doing it that way, however, you lose the flexibility that `MenuItem` provides. It isn't just a simple component, it also allows for `children` that would render a completely custom component while keeping the `MenuItem` logic. 

Also, `MenuItem's` aren't the only valid children, they are the only valid children that `Menu`  will attach listeners to. For component library i'd prefer to error on the side of more flexibility rather than less.

src/components/Menus/Menu.js
Eldwick: We don't have to go with this specific pattern, but I think it's very important we standardize these within `Snacks` to maintain standards and predictability.  FWIW I've used this pattern in large scale FE applications in the past and it allows you to create more abstract handlers if you have a know the shape of the payload.

src/components/Menus/DropdownMenu.js
Eldwick: How does `onRequestChange` play into that? Seems like you could still do that with just a `handleSelect` prop and an `open` prop.

src/components/Menus/Menu.js
Eldwick: Ahh I see, okay nice 👍

src/components/Menus/Menu.js
```diff
@@ -0,0 +1,204 @@
...
+  menuItemIsValid(menuItem) {
+    return menuItem && !menuItem.props.disabled
+  }
+
+  getMenuItemChildren() {
+    const { children } = this.props
+    const menuItemChildren = []
+
+    React.Children.map(children, (child) => {
+      if (child.type && child.type.displayName === 'MenuItem') {
```
Eldwick: Any chance we would want to wrap a `MenuItem` with a hoc? This would fail in that case.

dcocchia:
@jordanscode mentioned wanting an animation for the dropdown itself. Just want to tag him to make sure we start talking about it

jordanscode:
- The down arrow in the select component isn't vertically centered in the input
- Last night, @dcocchia and I had some talk about working more on animations. For the dropdown appearing and disappearing we should work on something! My idea:

On open:
1. Menu starts with a top offset higher than our intended placement and at 0% opacity
2. Slides quickly into place and animates opacity to 100%.

On close: Here we might only want opacity to fade out. Movement up could be odd, but movement downward could feel tactile. (needs experimentation)

dcocchia:
👍 Would love to see that animation, @jordanscode 

We may want to think about creating a separate animations section of the library eventually. But for now, let's try to make these animations somewhat extractable, @nbwar :)

nbwar:
Can we do the animations in V2 along with the mobile web version. This PR is already pretty large and it would be nice to get this release out so people can start using it.

dcocchia:
yeah, that's fair. Since we have a v2 planned already, I'm ok with that.

@Eldwick any remaining concerns about merging this one?

Eldwick:
Still have a few comments above but to summarize:
1. I think we should fail hard instead of rendering null for invalid elements
2. We should standardize the callback api
3. I think `onRequestChange` adds unnecessary complexity

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,207 @@
...
+      <div
+        onKeyDown={this.handleKeyDown}
+      >
+        <div style={{position: 'relative'}}>
+          { this.renderTriggerElement() }
+
+          <div
+            style={[
+              styles.menuContainer,
+              style,
```
Eldwick: I would expect a passed down `style` prop to apply to the top level html element. I'm imaging a case like:
```
<div style={{ display: flex }} >
  <div>Logo</div>
  <DropdownMenu style={{ flexGrow: 1 }} />
</div>
```

The above wouldn't render as I would expect and would be impossible to implement.

nbwar:
@Eldwick 
3. `onRequestChange` is needed to override default open/close of the dropdown. Not sure how you would get around this any other way without always having to manually open/close it, which adds complexity to using the library

5. What instance would we not want to prevent the default on an event in the dropdown trigger? And what other instances should we not by default prevent?

6. I originally wrote the Select trigger as it's own element however since it is wrapped in a radium enhancer, I lost the ability to have access to the refs and call .focus on the component. This is also why I need to pass the `focus` prop to each `MenuItem` rather than just calling `.focus()`

Eldwick:
I'm not following why `onRequestChange` is necessary if the component is controlled. How is this different than a standard controlled component whose state is controlled by the parent? Could you maybe give an example? I see it being controlled like:
```
class UnconrolledParent {
  render() {
    return (
      <div>
        <DropdownMenu />
      </div>
    )
  }
}

class ConrolledParent {

  handleClick = (e) => {
    this.setState(({ forceOpen, open }) => ({ open: forceOpen ? true : !open }))
  }
  
  render() {
    return (
      <div>
        <DropdownMenu onClick={this.handleClick} open={this.state.open}/>
      </div>
    )
  }
}
```

Eldwick:
We may not want to prevent default if the children of the elements was submitting a form or was a link to an external site or we wanted to rely on the browser for routing.

Eldwick:
I actually really like what you did with `MenuItem` seems like a more reactive approach to focusing the appropriate element. Maybe we should replicate that for the trigger element?

nbwar:
RE `onRequestChange`: The example you provided works only when `onClick` is responsible for controlling open/close. However, for this to work with `Dropdown` you would need to provide callbacks for `onBlur`, `onClick`, & `onSelect`. Plus i would also need to provide callbacks for `onKeyDown`  for `enter` and `escape`. It seemed cleaner to provide one callback `onRequestChange` that would notify the parent that the dropdown is attempting to open/close rather than knowing have to override all of these callback to achieve the same effect.

RE `preventDefault`: I understand that, however, if child element for dropdown is trying to submit a form or route to a different url while attempting to open the dropdown, the wrong element is being used. I don't see an instance where we would want to do any of those things and open a dropdown at the same time. It just makes it easer for consumers to not have to worry about preventing default if they were to use an `<a>` tag for trigger elements. However, `MenuItem` with a `<a>` tag we may want to leave `preventDefault` to the parent.

Eldwick:
Ahh I get it now for the `onRequestChange`, I think I was confused because of the naming. Is this it more like `onOpenToggle` or something? 

Sweet then maybe let's just leave out the `preventDefault` on the `handleKeyDown` handler in MenuItem.

src/components/Forms/Select.js
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  state = {
+    isOpen: false,
+    selectedOption: this.props.defaultOption || this.props.selectedOption || null
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { selectedOption } = nextProps
+    if (typeof selectedOption !== 'undefined' && selectedOption !== this.state.selectedOption) {
```
julianhr: this can be reduced to `if (selectedOption && selectedOption !== this.state.selectedOption) {`

src/components/Forms/Select.js
```diff
@@ -0,0 +1,355 @@
...
+    const {
+      disabled,
+      required,
+      hasError,
+      hintText,
+      floatingLabelText
+    } = this.props
+
+    const { isOpen, selectedOption, isFocused } = this.state
+    const hasValue = selectedOption !== null
```
julianhr: This constant is unnecessary. You can use `selectedOption` to verify if it has value. Also, this doesn't catch the case when selectedOption could be undefined.

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,211 @@
...
+  static defaultProps = {
+    open: null
+  }
+
+  state = {
+    open: false
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open !== null && this.props.open !== nextProps.open) {
```
julianhr: This can be reduced to `if (nextProps.open && this.props.open !== nextProps.open) {`

src/components/Menus/DropdownMenu.js
```diff
@@ -0,0 +1,211 @@
...
+    setTimeout( () => {
+      if (this.controlledOpen()) {
+        onRequestChange && onRequestChange(false)
+      } else {
+        this.close()
+      }
+    }, 100)
+  }
+
+  handleKeyDown = (event) => {
```
julianhr: To me this is a big accessibility concern @dcocchia. I suggest that we turn the selectable elements to either an anchor tag or a button. Take a look at these examples:

http://pauljadam.com/demos/rwd-aria-menubar.html
http://pauljadam.com/demos/aria-menubar.html

src/components/Menus/DropdownMenu.js
julianhr: I would favor prefixing with `on` instead of `handle`. Shorter and sweeter.

src/components/Menus/MenuItem.js
```diff
@@ -0,0 +1,207 @@
...
+      role,
+      style,
+      tabIndex,
+      useTabIndex
+    } = this.props
+
+    return (
+      <div
+        ref={(node) => this.menuItem = node}
+        role={role}
```
julianhr: I would be against this implementation. We should not be using a div. Please take a look at these two examples:

http://pauljadam.com/demos/rwd-aria-menubar.html
http://pauljadam.com/demos/aria-menubar.html

dcocchia:
Alright, holiday times past. Let's merge this thing! @nbwar @Eldwick @julianhr Any remaining concerns or work for this one?

nbwar:
@dcocchia This is ready to go on my part unless anyone else has any more concerns

julianhr:
@nickelser As discussed with @dcocchia I will test this out for Accessibility later today. If it behaves well I will approve this PR. Let's wait to merge this PR until then.

nickelser:
wrong nick, i think

julianhr:
Yep, sorry about that.

nbwar:
@julianhr 

1. The button doesn't have focus because the component `Button` has set focus to `outline: none` if this is a behavior we want to change it will change all Buttons. 

2. fixed

3. I'll ping you

PR #94, opened 2018-01-04
Logan/add button aria label

logtheman:
Add aria label props to the Button component for accessibility.
@julianhr

### DIFF
--- a/src/components/Buttons/Button.js
+++ b/src/components/Buttons/Button.js
@@ -173,6 +173,7 @@ const Button = props => {
     disabled: props.disabled,
     tabIndex: props.tabIndex,
     type: props.type,
+    'aria-label': props.aria.label,
     style: [
       baseStyles,
       sizeStyles[props.size],
@@ -222,6 +223,11 @@ const Button = props => {
 }
 
 Button.propTypes = {
+  /** Hash of Aria attibributes used by screen readers */
+  aria: PropTypes.shape({
+    label:PropTypes.string,
+  }),
+
   /** Size of the button. */
   size: PropTypes.oneOf(['tiny', 'small', 'standard', 'large']),
 
@@ -272,6 +278,7 @@ Button.propTypes = {
 }
 
 Button.defaultProps = {
+  aria: {},
   size: 'standard',
   snacksStyle: 'primary',
   disabled: false,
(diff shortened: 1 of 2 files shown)


dcocchia:
@julianhr Can you take a look at this one when you get a chance?

@logtheman Can you rebase with master to ensure we can merge when we're ready?

wfro:
This was why we originally included `elementAttributes`, for any other arbitrary attrs that a button might need. If we wanted to do this as just a convenience API, I'd vote for `ariaLabel` as a string.

logtheman:
@wfro - Maybe we need to revisit the implementation in other components and decide on a consist method. Radio Button implements using aria as an object with a label attribute, which is the pattern I followed here after discussing with @julianhr. However, Circle Button uses `ariaLabel` as you suggested. Right now we don't have consistency. We should agree on a standard.

wfro:
@logtheman yep agreed on consitency. Since `ariaLabel` is the most common thing that would be added to `elementAttributes`, that's my vote (and we can change the Radio api, and and it to the `Button` api.

dcocchia:
We're to the point with Snacks now that we can start hardening our API, but we'll need to agree on what it should be. There are a couple things we need to get consistency on that I think we should talk through. Like `style` vs `styles` prop. We also have been talking about standardizing our event handing system. I'll spin up a Snacks specific Slack room so we can start talking through these kinds of things.

julianhr:
My two cents is that aria attributes vary a lot depending on context. Therefore, a flexible system is needed. Doing a simple search for aria usage in our codebase yields this:

screen shot 2018-01-11 at 12 00 14 pm

aria-hidden
aria-describedby
aria-label

These were only the first top results. The reason I opted for the Shape is that the attributes are diverse and aria is always the prefix. Therefore, my vote would be for the shape. Agree that we need to come up with a consistent method of describing them.

wfro:
I'm gonna change my mind here and advocate for `elementAttributes`, it's original intention was to be a flexible way to add these kinds of html attributes. If we make sure the rest of the component's API has what it needs, I'm not too worried about `elementAttributes` being a junk drawer.

logtheman:
Closing in favor of spreading unspecified props rather than specifying a aria object.

julianhr:
@loganmurdock I think it would be better to use this PR for the changes you will make so that, when merged, if ever needed we can refer back to this conversation searching through the PR history. This has been helpful in the past for me.

PR #95, opened 2018-01-09
Add LoadingBox component

Eldwick:
Create `LoadingBox` component that fills the width of it's parent container by default and can be easily styled for specific use cases.


src/components/Loading/LoadingBox.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  static defaultProps = {
+    dark: false,
+    light: false,
+  }
+
+  componentWillReceiveProps({ dark, light }) {
+    if (dark && light) {
+      throw new TypeError('LoadingBox can not receive both dark and light props')
```
dcocchia: Rather than having two bools controlling the same aspect of the component, would it make sense to have one theme prop? 

ex
```js
<LoadingBox theme='dark' />
```
This would also closer match the api for the `Button` component: https://instacart.github.io/Snacks/#button

```js
<Button 
  snacksStyle='primary' 
  size='large' 
/>
```

src/components/Loading/LoadingBox.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  shouldComponentUpdate({ dark, light, style }) {
+    const { dark: oldDark, light: oldLight, style: oldStyle } = this.props
+    return dark !== oldDark || light !== oldLight || style !== oldStyle
+  }
```
dcocchia: Do we need this in a `PureComponent`? Isn't React already doing a shallow object check on the state?

Eldwick:
@dcocchia Added standard abstractions and improved the API:

dcocchia:
Looks good! If you remove the `/dist` changes, I'll go ahead and approve/merge :)

PR #97, opened 2018-01-11
Transitions

Eldwick:
Add Grow, Fade, and Slide transitions.

Fade:

Grow:

Slide:

Combo:

Select:

Dropdown:


src/components/Transitions/Grow.js
```diff
@@ -0,0 +1,157 @@
...
+
+  get initialStyles () {
+    const { start: startScale } = this.scale
+    const { transitionTime } = this.props
+
+    return {
+      maxHeight: 0,
+      overflow: 'hidden',
+      transform: `${this.transformAxis}(${startScale})`,
+      transition: `all ${transitionTime}ms ease-in-out`,
```
dcocchia: `ease-in-out` is a good default, but I think we should have a prop to override this to other timing settings, as well. For [`ScrollTrack`](https://instacart.github.io/Snacks/#scrolltrack) we call that prop `scrollTimingFunction`, so maybe `timingFunction` or something to keep it similar.

src/components/Transitions/Slide.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+
+  get initialStyles() {
+    const { transitionTime, width } = this.props
+
+    return {
+      width: width,
+      transform: `${this.transformAxis}(-${width}px)`,
+      transition: `all ${transitionTime}ms ease-in-out`,
```
dcocchia: Same here :)

src/components/Loading/docs/LoadingBox.md
```diff
@@ -1,4 +1,3 @@
-### LoadingBox example:
 Light:
```
dcocchia: was removing the above line intentional?

package.json
```diff
@@ -66,7 +66,8 @@
     "prop-types": ">=15",
     "radium": "^0.18.0",
     "react": ">=15",
-    "react-dom": ">=15"
+    "react-dom": ">=15",
+    "react-transition-group": "^2.2.1"
```
dcocchia: Need to add this to the `devDependencies` as well.

styleguide.config.js
```diff
@@ -50,6 +50,10 @@ module.exports = {
       name: 'Components',
       content: 'docs/components.md',
       sections: [
+        {
+          name: 'Transitions',
+          components: 'src/components/Transitions/[A-Z]*.js'
+        },
```
dcocchia: We should try to keep these alphabetical, I think

dcocchia:
@wfro can you be our second pair of eyes for this one?

wfro:
Looks great so far, I'll take another pass in more detail tomorrow morning

src/components/Loading/docs/LoadingBox.md
Eldwick: Yup, it was repetitive since styleguide adds a heading already

src/components/Transitions/Fade.js
```diff
@@ -0,0 +1,121 @@
...
+    timeout: TIMEOUT,
+    transitionTime: TRASITION_TIME,
+    opacity: OPACITY_DEFAULT,
+    axis: AXIS,
+    style: {},
+    in: true,
+    appear: true,
+  }
+
+  get opacity() {
```
wfro: Any particular reason for the getters? Not opposed to them at all, just curious :)

src/components/Transitions/Grow.js
wfro: Also a fan of flexibility, even if 99% of users would stick with the defaults. Maybe less important, but could also consider letting the transition property (`all`) be overriden via props

src/components/Transitions/Grow.js
```diff
@@ -0,0 +1,157 @@
...
+
+  renderChild = (state) => {
+    const { style, children } = this.props
+    const styles = {
+      ...style,
+      ...this.initialStyles,
+      ...this.transitionStyles[state]
+    }
+    return (
+      <div style={styles} >
```
wfro: Extra space

src/components/Transitions/Slide.js
```diff
@@ -0,0 +1,104 @@
...
+    transitionTime: TRASITION_TIME,
+    width: WIDTH,
+    axis: AXIS,
+    style: {},
+    in: true,
+    appear: true,
+  }
+
+  get transformAxis() {
+    return this.props.axis == 'x' ? 'translateX' : 'translateY'
```
wfro: Any reason to not use === here?

wfro:
Left a few comments, I'm fine merging this as is (or adding those few props additions)

wfro:
Playing around in the styleguide a bit, with slide for example is there a way to start the animation from the bottom up instead of top down? Similarly is it possible to go right to left instead of left to right?

Eldwick:
@Will I can add in an `invert` prop that will allow you to go right to left and bottom up.

will:
Nah that's ok

src/components/Transitions/Fade.js
Eldwick: I like the API and I think it encourages good separation of concerns

dcocchia:
@Eldwick I should have pointed you to @jordanscode's[ comments](https://github.com/instacart/Snacks/pull/91#issuecomment-350068118) on what he was hoping for with dropdown transitions:

```
On open:

Menu starts with a top offset higher than our intended placement and at 0% opacity
Slides quickly into place and animates opacity to 100%.

On close: 

Here we might only want opacity to fade out. 
Movement up could be odd, but movement downward could feel tactile. (needs experimentation)
```

Eldwick:
@dcocchia I kinda like the movement back up, feels like a drawer closing. But I can add in a Fade aswell

dcocchia:
I would give @jordanscode a run through of it to see what he thinks. I'm not too opinionated on it :)

PR #109, opened 2018-01-19
Animate both width and height for Grow

Eldwick:
`Grow` used to only animate height, which could introduce jank for components that are horizontal to the component. Updated to animate both max width and height axis but maintained the single axis transformation.


dcocchia:
@Eldwick this lgtm, but can you remove the docs files from the branch?

PR #114, opened 2018-01-27
Radio, RadioGroup and Checkbox compatible with Safari

julianhr:
## What
The components Radio, RadioGroup and Checkbox are now compatible with Safari

## Why
Before the background images for these components were not displaying in Safari

_**BEFORE**_
screen shot 2018-01-27 at 1 45 07 am

**_AFTER_**
screen shot 2018-01-27 at 1 45 18 am


src/base/RadioCheckboxBase.js
```diff
@@ -39,30 +40,26 @@ function imgValidator (props, propName) {
...
+      selected      : imgValidator,
       disabled      : imgValidator,
     }),
     btnType       : PropTypes.oneOf(['radio', 'checkbox']).isRequired,
     isDisabled    : PropTypes.bool,
     children      : PropTypes.string,
     id            : PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     isSelected    : PropTypes.bool,
-    onClick       : PropTypes.func,
+    onChange       : PropTypes.func,
```
julianhr: The reason for this prop change is that React kept complaining that when setting the `checked` attribute on the `input` tag, an `onChange` callback also had to be defined.

src/base/RadioCheckboxBase.js
```diff
@@ -88,40 +85,35 @@ class RadioCheckboxBase extends Component {
...
-          src={bkgImage}
-          tabIndex='-1'
-          alt=''
-        />
+        <BkgSvg />
         <input
           id={id}
           type={btnType}
-          onClick={this.onClick}
+          onChange={this.handleChange}
```
julianhr: The reason for this prop change is that React kept complaining that when setting the `checked` attribute on the `input` tag, an `onChange` callback also had to be defined.

src/base/RadioCheckboxBase.js
```diff
@@ -39,30 +40,26 @@ function imgValidator (props, propName) {
...
+      selected      : imgValidator,
       disabled      : imgValidator,
     }),
     btnType       : PropTypes.oneOf(['radio', 'checkbox']).isRequired,
     isDisabled    : PropTypes.bool,
     children      : PropTypes.string,
     id            : PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     isSelected    : PropTypes.bool,
-    onClick       : PropTypes.func,
+    onChange      : PropTypes.func,
```
julianhr: The reason for this prop change is that React kept complaining that when setting the checked attribute on the input tag, an onChange callback also had to be defined.

spencerschack:
This looks good to me except `bkgSvg`. No reason to limit the type of that prop to svg only, it can be any react component. Maybe something like `imageElements`? Or break it into separate props like `selectedElement` or `selectedIndicator`.

julianhr:
As discussed with @spencerschack, it probably would be better to leave the implementation open to pass any image type for the background images. However, that would require creating a new component to handle inline SVG's vs img tags. Since the odds of using img tags for these components is very small (we were and that's what I'm changin here for compatibility purposes) I think we might over-engineer the component for something we may not use.

If @dcocchia @dcocchia agree, it'd be great if either of you can approve this PR so I can merge it. Without it, I can't.

PR #121, opened 2018-02-28
Limit files pulled down when installed

tmarshall:
Using [npm's files field](https://docs.npmjs.com/files/package.json#files) to limit what is pulled down when installing.

To test, you can install in a local dir via `npm i git@github.com:tmarshall/Snacks.git#tmarshall/files --save` (and then dig into the files added to `node_modules`)


package.json
```diff
@@ -26,6 +26,12 @@
     "components",
     "component library"
   ],
+  "files": [
+    "dist",
+    "CONTRIBUTING.MD",
```
tmarshall: kept this in in case an engineer digs into the `node_modules` - though may be unnecessary

tmarshall:
old. closing.

PR #125, opened 2018-03-10
[Button] Use active styles for focus

wfro:
Adds the correct styles for focus states (same as active).


dcocchia:
@wfro can you rebase this one with master? @spencerschack added the flat button styles, so might need a little massaging :)

wfro:
@dcocchia should be good to go. Took a slightly different approach than last round but tested and everything looks good.

PR #132, opened 2018-03-15
Dropdown hidden elements clickable fixes #129

### LINKED ISSUE #129: Dropdown menu items are actionable when hidden
When a `Select` with `MenuItems` is used on something like a modal, where the z-index hierarchy is removed/higher than the parent page, the dropdown menu items can be interacted with even when hidden.

kapture 2018-03-14 at 13 56 45

nbwar:
Turning off pointer events for dropdown elements that should be hidden

@dcocchia 
@Eldwick


dcocchia:
This seems reasonable to me. Between the `pointer-events: none`, `tab-index=-1` and `aria-hidden:  true`, I think we should be good.

How do you feel about this solution @Eldwick?

Eldwick:
LGTM, nice catch @nbwar

PR #133, opened 2018-03-15
[WIP] SVG icons

spencerschack:
screenshot

### What
Upside-down SVG icons

### Why
🙃


src/components/Icon/Icon.js
```diff
@@ -1,55 +1,34 @@
...
-const Icon = props => {
-  const { style, onClick } = props
-  const icon = getIcon(props)
+function Icon({ name, style, onClick }) {
+  const IconRenderer = icons[name]
   return (
-    <i
-      style={[baseStyles, style]}
+    <IconRenderer
+      style={{...baseStyles, ...style}}
```
spencerschack: The styles were not applied to the svg with the array syntax 🤷‍♂️

src/components/Icon/icons.js
```diff
@@ -0,0 +1,171 @@
+export {default as account} from './icons/account.svg'
```
spencerschack: If anyone knows a better way to do this, please let me know

Issue #117, opened 2018-02-12
Export themePropTypes for external propTypes validations


dcocchia:
Currently users have to impliment their own propType validation for the `snacksTheme` prop, which is guaranteed to break over time. If we export our internal util, `themePropTypes`, users could do

```js
import { Component } from 'react'
import { withTheme, themePropTypes } from 'ic-snacks'

class MyComponent extends Component {
  static propTypes = {
    snacksTheme: themePropTypes
  }

  render(...)
}
```

This is less code for them, and as they update their library version, the theme validation would be updated with it.

dcocchia:
Done in https://github.com/instacart/Snacks/pull/134

PR #134, opened 2018-03-16
Export themePropTypes

dcocchia:
Snacks users who want to theme their own components using the `withTheme` HOC currently must write out their own propTypes for the `snacksTheme`. This is repetitive, but also could be fragile. If the theme config changes, users shouldn't have to re-write all their propTypes.

Now they can do
```js
import { withTheme, themePropTypes } from 'ic-snacks'

const MyComponent = props => {
  return (
    <p style={{ color: props.snacksTheme.color.action }}>
      Hi!
    </p>
  )
}

MyComponent.propTypes = {
  snacksTheme: themePropTypes
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,23 +33,30 @@ import withTheme from './styles/themer/withTheme'
 import Slide from './components/Transitions/Slide'
 import Grow from './components/Transitions/Grow'
 import Fade from './components/Transitions/Fade'
+import { themePropTypes } from './styles/themer/utils'
 
 export {
-  // base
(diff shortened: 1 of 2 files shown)


dcocchia:
Fixes https://github.com/instacart/Snacks/issues/117

dcocchia:
For now I think keeping the name generic is ok. Since we plan/hope to move out the themer/theming stuff into it's own lib, I'm going to keep the Snacks name out of it until we have a better idea of what that will look like.

PR #140, opened 2018-04-06
New PhoneNumberField component

dcocchia:
### What
- Adding new PhoneNumberField component, based on our `TextField` component
- Pulled in `react-text-mask` for the input masking
- Pulled in `enzyme-to-json` so I could test a snapshot of a fully mounted component - the react-text-mask component uses a render callback to work, so needed this
- Fixes alignment issues with our `HintText` component

kapture 2018-04-06 at 16 03 32


src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+import Radium             from 'radium'
+import { colors }         from '../../styles'
+import FormComponent      from './FormComponent'
+import ValidationError    from './ValidationError'
+import FloatingLabel      from './FloatingLabel'
+import TextFieldHint      from './TextFieldHint'
+import ServerError        from './ServerError'
+import HelperText         from './HelperText'
+import withTheme          from '../../styles/themer/withTheme'
+import MaskedTextInput    from 'react-text-mask'
```
wfro: Very nitpicky but package/relative imports are usually at the top

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+            onBlur={this.handleInputBlur}
+            onChange={this.handleInputChange}
+            onFocus={this.handleInputFocus}
+            onKeyDown={this.handleKeyDown}
+            autoComplete={autoComplete}
+            placeholder=""
+            defaultValue={value !== undefined ? undefined : defaultValue}
+            disabled={disabled}
+            keepCharPositions={true}
+            render={(ref, props) => (
```
wfro: The fabled render prop :D

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+              <input
+                ref={(input) => {
```
wfro: Our form system still requires us to have refs right? In a perfect world I think it'd be nice to only have controlled inputs, and we can have "controllable" input components. Where the component itself manages it's own state, but it also gives the caller the option to pass `onChange` and `value`.

src/components/Forms/__tests__/PhoneNumberField.spec.js
```diff
@@ -0,0 +1,122 @@
...
+          hintText="(555) 555-555"
+          onChange={onChange}
+        />
+      </div>
+    </StyleRoot>
+  )
+
+  wrapper.find('input').simulate('change', {target: {value: 'My new value'}})
+
+  expect(onChange.calledOnce).toBe(true)
```
wfro: Would it make sense to actually test the callback gets the right argument here?

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+
+  static defaultProps = {
+    autoComplete: 'on',
+    disabled    : false,
+    defaultValue: null,
+    onKeyDown: () => {} // eslint-disable-line no-empty-function
+  }
+
+  state = {
+    hasValue: this.props.defaultValue !== null || Boolean(this.props.value)
```
wfro: Is it worth it to look into trying to consolidate some of this with the existing text input? They do a few similar things (like this default state, similar blur/focus/input change logic, etc.). Maybe not as part of this PR but in general we can look into extracting some core piece we could use in both.

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  handleInputFocus = (e) => {
+    this.setState({isFocused: true})
+    this.props.onFocus && this.props.onFocus(e)
+  }
+
+  handleInputBlur = (e) => {
+    this.setState({isFocused: false})
+    this.props.onBlur && this.props.onBlur(e)
```
wfro: nit - mostly personal preference but I like giving noops as a default value for optional callback props like this (which means lightly less code as you don't have to check that they exist).

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+      autoComplete,
+      snacksTheme
+    } = this.props
+
+    const {
+      hasValue,
+      isFocused
+    } = this.state
+
+    const inputId = id
```
wfro: irrc you can give a destructured prop a new name inline:

```
const {
  id: inputId
} = this.props
```

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+import ValidationError    from './ValidationError'
+import FloatingLabel      from './FloatingLabel'
+import TextFieldHint      from './TextFieldHint'
+import ServerError        from './ServerError'
+import HelperText         from './HelperText'
+import withTheme          from '../../styles/themer/withTheme'
+import MaskedTextInput    from 'react-text-mask'
+import { themePropTypes } from '../../styles/themer/utils'
+
+const styles = {
```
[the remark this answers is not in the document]
dcocchia: Just the prop that gets passed in. Internal variables are sort of up in the air

src/components/Forms/PhoneNumberField.js
```diff
@@ -0,0 +1,300 @@
...
+          />
+
+          <TextFieldHint
+            text={'(555) 555-555'}
+            show={!hasValue && isFocused}
+            disabled={disabled}
+          />
+
+          <MaskedTextInput
+            mask={['(', /[1-9]/, /\d/, /\d/, ')', ' ', /\d/, /\d/, /\d/, '-', /\d/, /\d/, /\d/, /\d/]}
```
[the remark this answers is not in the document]
dcocchia: In the future, yeah. But for now this is US only

src/components/Forms/PhoneNumberField.js
dcocchia: Yeah, I agree. I imagine there will be at least one or two other inputs we'll add that could share this logic as well

src/components/Forms/__tests__/PhoneNumberField.spec.js
```diff
@@ -0,0 +1,122 @@
...
+    </StyleRoot>
+  )
+
+  wrapper.find('input').simulate('focus')
+  
+  expect(toJson(wrapper)).toMatchSnapshot()
+})
+
+it('fires the onFocus prop', () => {
+  const onFocus = spy()
```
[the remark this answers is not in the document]
dcocchia: Would be nice to rely on one fewer dependencies for testing :) I'll give it a shot here and see how it goes

src/components/Forms/PhoneNumberField.js
dcocchia: Yeah, the `Form` component requires the `input` ref right now. You could use this component outside of a `Form`, and pass your own `onChange` handler down. Basically you'd implement what the `Form` does now?

src/components/Forms/PhoneNumberField.js
dcocchia: In all it's glory. It was super helpful in this situation :D

PR #146, opened 2018-04-19
Flat Button style

### LINKED ISSUE #131: Link style for Button component - "Flat"
It's common for products to have a button styled like a link. We often use anchor tags or the `Link` component to do this. This has issues with styles (font is too small, hover state has underline) and accessibility (links should navigate, buttons should cause actions).

Let's add a `flat` variant to the `snacksStyle` prop for our `Button` component. The styles will be exactly the same a the `secondary` options, just with no border. This includes the padding, hover states, etc. So not exactly the same as a link, but close.

cc @jordanscode

spencerschack:
Buttons without borders!

buttnos

Fixes #131


wfro:
Nice! My only comment would be the API, I think `snacksStyle="link"` is more intuitive

spencerschack:
That's what I started with but I just followed Dom's suggestion in #131. I think "flat" may be a little clearer because "link" might make it seem like it would render as a link possibly with underline styles that you would expect with a link.

dcocchia:
"Flat" is also what we plan to call it across platforms, which don't have "links" :)

PR #153, opened 2018-05-17
Responsive helpers

wfro:
### What

This commit adds a few responsive helpers (inspired by bootstrap 4 https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints)

* `up`
* `down`
* `between`
* `only`

### Why

Currently the `responsive` utility only allows users to target a single screen size at a time. So if a user wants to apply styles to xs, sm, and md screens, they'd have to write something like:

```
const styles = {
  [responsive.xs]: { color: 'blanchedalmond' },
  [responsive.sm]: { color: 'blanchedalmond' },
  [responsive.md]: { color: 'blanchedalmond' }
}
```

These new utils provide a much more flexible API, and now the same effect can be achieved with:

```
const styles = {
  [responsive.down('md')]: { color: 'blanchedalmond' }
}
```

### DIFF
--- a/src/styles/__tests__/responsive.spec.js
+++ b/src/styles/__tests__/responsive.spec.js
@@ -1,0 +1,51 @@
+import responsive, { sizes } from '../responsive'
+const { __v2__ } = responsive
+
+const { up, down } = __v2__
+
+describe('up', () => {
+  it('ignores xs as an argument because it has no lower bound')
+
+  it('returns a media query that matches screens larger than the given screen size', () => {
+    const ss = Object.keys(sizes)
+
+    // TODO: xs isn't valid for up, it has no lower bound
+    ss.filter(size => size !== 'xs').forEach(size => {
+      expect(up(size)).toEqual(`@media (min-width: ${sizes[size].min}px)`)
+    })
+
+    // TODO: should maybe do one sanity check in case i messed up attaching the constsnts in sizes?
+    // should i import constants or hardcode
+    expect(up('sm')).toEqual(`@media (min-width: ${768}px)`)
+  })
+})
+
+describe('down', () => {
+  it('ignores xl as an argument because it has no upper bound')
+
+  it('returns a media query that matches screens smaller than the given screen size', () => {
+    const ss = Object.keys(sizes)
+
+    // TODO: xl isn't valid for down, xl has no upper bound
+    ss.filter(size => size !== 'xl').forEach(size => {
+      expect(down(size)).toEqual(`@media (max-width: ${sizes[size].max}px)`)
+    })
+
+    // TODO: same as above
+  })
+})
+
+describe('only', () => {
+  it('works for xs')
+  it('works for xl')
+
+  it('works for every other size that has both a lower/upper bound', () => {
+    expect(1).toEqual(1)
+  })
+})
+
+describe('between', () => {
+  it("returns a media query that matches screens between each given size's min value", () => {
+    expect(1).toEqual(1)
+  })
+})
--- a/src/styles/__tests__/responsive.spec.js
+++ b/src/styles/__tests__/responsive.spec.js
@@ -1,49 +1,70 @@
-import responsive, { sizes } from '../responsive'
+import responsive, { breakpoints } from '../responsive'
 
 const { up, down, only, between } = responsive
 
 describe('up', () => {
-  it('ignores xs as an argument because it has no lower bound')
+  it('returns a media query that matches the given screen size or larger', () => {
+    const breakpointNames = Object.keys(breakpoints)
 
-  it('returns a media query that matches screens larger than the given screen size', () => {
-    const ss = Object.keys(sizes)
-
-    // TODO: xs isn't valid for up, it has no lower bound
-    ss.filter(size => size !== 'xs').forEach(size => {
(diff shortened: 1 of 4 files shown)


src/styles/responsive.js
```diff
@@ -1,30 +1,102 @@
...
 const SCREEN_XS_MAX = SCREEN_SM_MIN - 1   // 767
 const SCREEN_SM_MAX = SCREEN_MD_MIN - 1   // 831
 const SCREEN_MD_MAX = SCREEN_MDLG_MIN - 1 // 1039
 const SCREEN_MDLG_MAX = SCREEN_LG_MIN - 1 // 1247
 const SCREEN_LG_MAX = SCREEN_XL_MIN - 1   // 1455
 
+// TODO: exporting just for test (so far at least), smell?
+export const sizes = {
+  xs: {
+    // TODO: no min, hmmm
```
dcocchia: maybe 0? Does having a `min` help?

src/styles/responsive.js
```diff
@@ -1,30 +1,102 @@
...
+    min: SCREEN_MDLG_MIN,
+    max: SCREEN_MDLG_MAX
+  },
+  lg: {
+    min: SCREEN_LG_MIN,
+    max: SCREEN_LG_MAX
+  },
+  xl: {
+    min: SCREEN_XL_MIN
+    // TODO: no max, hmmm
```
dcocchia: If a `max` is helpful, `Number.POSITIVE_INFINITY`?

src/styles/responsive.js
```diff
@@ -1,30 +1,102 @@
...
     mdLg: SCREEN_MDLG_MIN,
     lg: SCREEN_LG_MIN,
-    xl: SCREEN_XL_MIN,
-  }
+    xl: SCREEN_XL_MIN
+  },
+
+  // Helpers
+  up,
+  down,
```
michalDob: I find up and down a bit head scratchy as names, would below, above be better ?

spencerschack: Agreed. But above and below don’t seem inclusive of the passed in size...

spencerschack: `gte` and `lte` would be unambiguous

michalDob: I like that but not a fun of the abbreviations could do aboveAnd('md')/belowAnd('sm')? or greaterOrEqual('md) lessOrEqual('sm')

src/styles/__tests__/responsive.spec.js
```diff
@@ -0,0 +1,49 @@
+import responsive, { sizes } from '../responsive'
+
+const { up, down, only, between } = responsive
+
+describe('up', () => {
+  it('ignores xs as an argument because it has no lower bound')
+
+  it('returns a media query that matches screens larger than the given screen size', () => {
```
michalDob: larger than or equal ?

src/styles/__tests__/responsive.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+    // TODO: should maybe do one sanity check in case i messed up attaching the constsnts in sizes?
+    expect(up('sm')).toEqual(`@media (min-width: ${768}px)`)
+  })
+})
+
+describe('down', () => {
+  it('ignores xl as an argument because it has no upper bound')
+
+  it('returns a media query that matches screens smaller than the given screen size', () => {
```
michalDob: smaller than or equal to?

src/styles/responsive.js
wfro: 👍 thank, I went with this. Having a `min` helps avoid having to add a few special cases.

wfro:
Thanks for the comments all - this is ready for another round of review.

As far as function names, I personally don't have an issue with up/down. If we get consensus on something else I'm happy to update. IMO it's difficult to convey inclusion without getting wordy (3+ words). And abbreviations might negate what I see as the benefit to a different name - being explicit enough to not have to double check the docs for specific behavior.

src/styles/responsive.js
```diff
@@ -1,30 +1,113 @@
...
+  },
+  xl: {
+    min: SCREEN_XL_MIN
+  }
+}
+
+const assertValidSizes = (...sizes) => {
+  sizes.forEach(size => {
+    if (breakpoints[size] === undefined) {
+      throw new Error(`Screen size(s) ${sizes.join(', ')} not supported. Must be one of ${Object.keys(breakpoints).join(', ')}`) // eslint-disable-line
```
dcocchia: nice! We should probably do more of this kind of thing to help out the users of this library.

Looks like I need to edit the eslint rule for this to disregard strings

styleguide.config.js
```diff
@@ -20,12 +20,21 @@ module.exports = {
       content: 'docs/installation.md'
     },
     {
-      name: 'Colors',
-      content: 'docs/colors.md'
-    },
-    {
-      name: 'Spacing',
-      content: 'docs/spacing.md'
+      name: 'Utilities',
```
dcocchia: good choice - We should likely do more fo this sub-sectioning :)

src/styles/responsive.js
wfro: Yeah, I figured this was better than returning a broken media query. One thing I was curious about was whether I could/should make this dev-only 🤔

Issue #130, opened 2018-03-14
Convert font icon to use SVGs


dcocchia:
Our `Icon` component currently relies on an icon font. This has many downsides for both end users of products and for developers implementing Snacks on their products. Our designers would like to use SVG icons for consistency, as well.

We'll maintain the same api for the `Icon` component as we do now with the exception of the `code` prop. The `code` prop will need to be removed, since SVGs do not use hexadecimal. That was never really a great idea in the first place, so best to kill it now before we do a more final release.

Should fix https://github.com/instacart/Snacks/issues/120
Will also fix the `String.fromCodePoint` issue from https://github.com/instacart/Snacks/issues/76

dcocchia:
We'll also need to figure out exporting SVGs for use in other code bases. Not sure how tricky that will be 🤔

spencerschack:
My thoughts were to use a loader that converted SVGs to react components.

dcocchia:
That would solve a ton of issues, yeah. Maybe something like this lib https://github.com/smooth-code/svgr

spencerschack:
Wow that is impressive

dcocchia:
Fixed with https://github.com/instacart/Snacks/pull/155

PR #155, opened 2018-05-18
SVG Icons

spencerschack:
SVG Icons

Fixes #133


icon_exporter.rb
```diff
@@ -0,0 +1,57 @@
+#! /usr/bin/env ruby
```
spencerschack: @dcocchia This will probably come in handy in the future, where should we put it?

spencerschack:
So there are a lot of icon name changes here. I tried to match the design system as best as possible, but there are probably some icons that we should not rename. Would love feedback: @dcocchia @wfro @Eldwick

spencerschack:
We also need to decide what the API for sizing the icons will be. Now it's just `style={{fontSize: X}}` but that won't work anymore. We may want to restrict to keywords `size="small"` so that we have consistent icon sizes.

dcocchia:
@spencerschack this is great!

For circle ci issue - it looks like this started just on this branch 🤔. Is it possible that the svg icon process is taking up a ton of memory or something?

For sizing, totally agree that we should have a prop that accepts string. Much like our buttons. Let's work with design to get those standardized across 👍

tmarshall:
@spencerschack are the tests running in parallel? Until we figure out mem issues, you could try to force the svg tests to run in a series

spencerschack:
Excellent call @tmarshall!!! I added `--runInBand` to the `test` script and it fixed the OOM issue. The tests do not run too much longer in dev. Tests are now passing!

dcocchia:
Fixes https://github.com/instacart/Snacks/issues/130

PR #159, opened 2018-05-30
Tooltip v1

nbwar:
Shifts over if won't fit on screen when placement is `top` or `bottom`


src/components/Tooltip/TooltipPosition.js
```diff
@@ -0,0 +1,164 @@
...
+      offsetEl = offsetEl.offsetParent
+    }
+
+    return offset
+  }
+
+  calculatePosition() {
+    const { placement } = this.props
+    const target        = this.getTarget()
+    const overlay       = findDOMNode(this)
```
dcocchia: Let's give our `div` below a `ref` and then we don't need to use `findDOMNode`

src/components/Tooltip/TooltipPosition.js
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  componentDidMount() {
+    this.calculatePosition()
+  }
+
+  getTarget = () => {
+    const { target } = this.props
+    const targetNode = typeof target === 'function' ? target() : target
+    return targetNode && findDOMNode(targetNode) || null
```
dcocchia: Is it possible to pass the node itself as the prop here so that we don't need to call `findDOMNode`?

src/components/Tooltip/TooltipPosition.js
```diff
@@ -0,0 +1,164 @@
...
+      if (placement === 'right') {
+        overlayRect.left = targetRect.left + targetRect.width + (SPACING * 2)
+        arrowPosition.left = -ARROW_SPACING
+      } else {
+        overlayRect.left = targetRect.left - overlayRect.width - (SPACING * 2)
+        arrowPosition.left = overlayRect.width - ARROW_SPACING - (ARROW_BORDER_SPACING / 2)
+      }
+    }
+
+    this.setState({overlayRect: overlayRect, arrowPosition: arrowPosition})
```
dcocchia: nit - can be a bit more concise 
```js
this.setState({ overlayRect, arrowPosition })
```

src/components/Tooltip/TooltipPosition.js
```diff
@@ -0,0 +1,164 @@
...
+    return (
+      <div style={computedStyles}>
+        { child }
+      </div>
+    )
+
+  }
+}
+
+export default TooltipPosition
```
dcocchia: This component is great because it abstracts out the complexity of positioning into a single space. I'm struggling to think of ways that make this less complex without just parting out the `calculatePosition` method into sub-methods. Maybe one method for top/bottom positioning and another for left/right would be good.

nbwar: Ya was thinking about how to make it less complex as well but a lot of variables are shared. When i work on v2 (paragraph tooltips) there will be added logic for left/right positioning for top. Can refactor then if we can wait.

src/components/Tooltip/calculateOffset.js
```diff
@@ -0,0 +1,42 @@
+// const OFFSET_TOP = {
```
Eldwick: Is this file no longer needed?

Eldwick:
Nice! Lookin good!

src/components/Tooltip/TooltipRootClose.js
```diff
@@ -0,0 +1,82 @@
...
+
+  componentDidMount() {
+    const { rootCloseEnabled } = this.props
+
+    if (rootCloseEnabled) {
+      this.addEventListeners()
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
```
Eldwick: `componentWillReceiveProps` is getting deprecated in react 17: https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/update/component_will_receive_props.html

src/components/Portal/Portal.js
```diff
@@ -0,0 +1,44 @@
...
+      PropTypes.func
+    ])
+  }
+
+  constructor(props) {
+    super(props)
+    this.setContainer(this.props.container)
+  }
+
+  componentWillReceiveProps(nextProps) {
```
Eldwick: Can we refactor this to avoid `componentWillReceiveProps` since it's getting deprecated?

src/components/Tooltip/TooltipRootClose.js
nbwar: Looks like the version of React we are using doesn't have the replacement method `static getDerivedStateFromProps` would need to update React first

PR #163, opened 2018-06-14
Release scripts

dcocchia:
This creates two new scripts for use when building and releasing a new version of Snacks.
kapture 2018-06-13 at 17 48 29

Note: video shows tests running twice. This is an odd consequence of switching to the master branch and running npm commands. Once this branch is merged, that issue should go away.


CONTRIBUTING.MD
```diff
@@ -141,3 +141,10 @@ or
 ```
 npm run styleguide:build
 ```
+
+## Releasing a new version of Snacks
+1. `npm run build_release`
```
tmarshall: why not `release:build` and `release:publish` to be consistent w/ `test:lint`?

scripts/release/build_release.js
```diff
@@ -0,0 +1,88 @@
...
+    return false
+  }
+
+  return true
+}
+
+const confirmVersion = {
+  name: 'Is this a patch, minor or major version?',
+  type: 'string',
+  pattern: /patch|Patch|minor|Minor|Major|major|/,
```
tmarshall: looks like an extra `|` was added at the end

tmarshall: also may be worth adding `^` and `$`

something like `/^\s*(?:[pP]atch|[mM]inor|[mM]ajor)\s*$/`

tmarshall: or just `/^\s*(?:major|minor|patch)\s*$/i`

CONTRIBUTING.MD
dcocchia: Good point 👍

apiv:
Nice!

PR #173, opened 2018-06-20
Use svgr over sprite loader for radio/checkbox

wfro:
This commit (so far):
* Removes the svg sprite loader. `svgr` seems to be working well for `Icon`, and the sprite sheet seemed to be the cause of most of the issues around getting `Radio`/`Checkbox` into the app. Win/win.
* Adjusts the `renderSvg` code to use svg markup directly as a component (similar to `Icon`)
* Adds `withTheme` to `Radio`/`Checkbox` (and updates the svgs to allow `fill` to be set via props)
* Generates new snapshots

Let me know if I'm headed in the right direction here. 

### TODO
- [x] Themes
- [x] Fix tests
- [x] Fix IE11 fill colors


webpack.config.js
```diff
@@ -1,13 +1,19 @@
...
 
 module.exports = {
   mode: 'development',
   module: {
     rules: [
       { test: /\.js$/, loaders: ['babel-loader'], exclude: /node_modules/ },
-      { test: /assets\/\w+\.svg$/, loader: 'svg-sprite-loader', exclude: /node_modules/, options: { extract: true } },
-      { test: /SVGIcon\/icons\/\w+\.svg$/, loader: 'svgr/webpack', exclude: /node_modules/, options: { extract: true } },
+      {
+        test: [
```
spencerschack: I updated this from `/\.svg/` because I had to specify different loaders, I think we can combine them again

wfro:
Got themes working + regenerated snapshots so this is ready for review

dcocchia:
awesome, thanks for seeing this through! Did you get a chance to try this out on IE11? We've had catastrophic svg issues on there previously, so want to make sure :)

wfro:
@dcocchia yeah, there's one (maybe two) ie11 issue(s) i'll work through today:

1. Fill color is always black (which is no good)
2. There's some slight jank when you select a radio - it fills the entire circle and then snaps to the correct size:



Definitely going to fix 1, curious whether you all think 2 warrants a fix (I could go either way at the moment).

PR #179, opened 2018-06-22
Remove tinycolor and add a shadeColor utility

wfro:
Themes define colors for hover states, but not focus states. To get focus colors we take the hover color and darken it by 20%. Previously we used `tinycolor` to do that, but it has a bunch of functionality we don't need. This replaces it with a single `shadeColor` function which gets us the same end result and shaves ~16k (not huge, but still nice) from the final bundle (uncompressed).

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,6 @@
     "react-text-mask": "^5.3.0",
     "semver": "^5.5.0",
     "svgr": "^1.9.2",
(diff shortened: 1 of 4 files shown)


dcocchia:
@wfro - Can you pick this one back up when you get some time? Looks like we should also add tests for these utils :)

PR #190, opened 2018-07-06
Stateless components cannot have refs in React 16

apiv:
Fixes this warning when used with react 16:
image

### DIFF
--- a/src/styles/themer/withTheme.js
+++ b/src/styles/themer/withTheme.js
@@ -5,7 +5,7 @@ import themer from './index'
 import { themePropTypes } from './utils'
 
 const isStateless = (Component) => {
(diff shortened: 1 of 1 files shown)


src/styles/themer/withTheme.js
```diff
@@ -4,6 +4,10 @@ import React, { Component } from 'react'
 import themer from './index'
 import { themePropTypes } from './utils'
 
+const isStateless = (Component) => {
+  return Component.prototype && !Component.prototype.render
```
spencerschack: I don't think we need to check the presence of `prototype`

apiv: True

PR #193, opened 2018-07-27
0.0.86

wfro:
* Allows users of `withTheme` to pass custom theme styles (#191)
* `withTheme` will no longer attempt to assign a ref to stateless components (#190)


wfro:
Ah good catch, updated (also linked to PRs).

PR #194, opened 2018-08-02
[tooltip] add programatic show/hide capability

scottjason:
Added a isVisible prop (If you don't pass anything it just works as it does now (sets to false on render)).

cc: @dcocchia @nbwar @spencerschack

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -21,7 +21,8 @@ class Tooltip extends PureComponent {
     target: PropTypes.node.isRequired,
     snacksStyle: PropTypes.oneOf(['primary', 'secondary', 'dark']),
     onDismiss: PropTypes.func,
-    onShow: PropTypes.func
+    onShow: PropTypes.func,
+    isVisible: PropTypes.bool,
   }
 
   static defaultProps = {
@@ -29,7 +30,8 @@ class Tooltip extends PureComponent {
     placement: 'bottom',
     size: 'small',
     onShow: noop,
-    onDismiss: noop
+    onDismiss: noop,
+    isVisible: false,
   }
 
   state = {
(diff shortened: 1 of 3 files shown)


scottjason:
It gets a little confusing to sort of have this both ways (have two different variables controlling the visibility). So right now w the pr as is, it's being instructed by the `isVisible` prop when `true` and then by the `state.show` when `false`.

spencerschack:
I would add a test for the `isVisible` prop and instead of adding another instance of the tooltip in the example, I would add a set of radio buttons that control `isVisible`

scottjason:
@spencerschack ok.

Also, what do you think about the fact that this is being controlled both by the `isVisible` prop which is coming from outside the component as well as the internal state of the component `isShow`?

spencerschack:
I think that's what Dom and I were originally thinking this would work, not sure how else it would be done.

src/components/Tooltip/Tooltip.js
```diff
@@ -70,11 +70,13 @@ class Tooltip extends PureComponent {
...
 
   render() {
+
     const {
       children,
       placement,
       size,
-      snacksStyle
+      snacksStyle,
+      isVisible,
```
dcocchia: be sure to add this to the `propTypes`

scottjason: 👍

scottjason:
@spencerschack Ok, yeah I think it should be fine. My other thought was to pass in the `isVisible` prop from the outside (or not) and have that define `state.show` (if and only if it's defined, not falsey).

But. there's problems w that i think because you'd have to set `this.state.show` in the render method.

scottjason:
@spencerschack Ok, now sharing the same tooltip (although I really don't think this makes any sense...this button is setup already to handle things using the internal state of the Tooltip component and now we're kind of blending in the other).

scottjason:
@spencerschack Ok I updated the PR and got rid of the extra example all together and now all of the examples are using the `isVisible` propr.

What i was originally trying to add as an example was showing how to use say an actual radio button as the target (like what we need to do in low stock) instead of having to click the radio button and then the toggle button.

PR #196, opened 2018-08-10
allow snacksTheme prop to be overridden

Brantron:
This PR will:
1. respect `snacksTheme` overrides
2. Throw errors in dev if the prop is overridden with anything other than an object.

Bonus - also some prettier changes to the eslintrc

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,6 +4,9 @@ module.exports = {
     "es6": true,
     "jest": true
   },
+  "globals": {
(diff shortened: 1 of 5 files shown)


dcocchia:
@Brantron we have a theme config validation util that might be a bit more encompassing than checking for if the theme is an object: https://github.com/instacart/Snacks/blob/master/src/styles/themer/utils.js#L52

Can we use that to verify theme overrides?

PR #197, opened 2018-08-13
Allow for Tooltip to become a controlled element

michalDob:
Change the behavior of controlled Tooltip.

Currently the tooltip is shown if either isVisible is true or internal state is true. And a click controls internal state. This leads to a behavior where even if isVisible is set to false the Tooltip will show up on click.

I think if a Tooltip is controlled, it should not override its visibility flag with internal state.

Behavior after this change:
If isVisible is undefined or null, no change in behavior.
if isVisible is set to false or true it fully controls the visibility of the tooltip.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ jspm_packages
 .yarn-integrity
 
 .DS_Store
+
+# IDE
+.idea
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -31,7 +31,6 @@ class Tooltip extends PureComponent {
     size: 'small',
     onShow: noop,
     onDismiss: noop,
-    isVisible: false,
   }
 
   state = {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -47,3 +47,6 @@ jspm_packages
 .yarn-integrity
 
 .DS_Store
+
+# IDE
+.idea
```
michalDob: for rubyminers out there

src/components/Tooltip/Tooltip.js
```diff
@@ -56,18 +55,19 @@ class Tooltip extends PureComponent {
...
+    const extraProps = isVisible == null ? { onClick: this.handleToggle.bind(this) } : {}
 
     return React.cloneElement(target, {
       ref: (node) => {
         this.trigger = node
       },
-      onClick: this.handleToggle.bind(this),
       'aria-haspopup': true,
-      'aria-expanded': show
+      'aria-expanded': isVisible || show,
```
michalDob: pretty sure this was just a bug

src/components/Tooltip/Tooltip.js
```diff
@@ -56,18 +55,19 @@ class Tooltip extends PureComponent {
   }
 
   renderTriggerElement() {
-    const { target } = this.props
+    const { target, isVisible } = this.props
     const { show } = this.state
 
     if (!target) { return }
+    const extraProps = isVisible == null ? { onClick: this.handleToggle.bind(this) } : {}
```
michalDob: if a component is controlled I dont think it should also control itself, onClick behavior should work only if isVisible is not passed in.

src/components/Tooltip/__tests__/Tooltip.spec.js
```diff
@@ -66,73 +66,51 @@ describe('Tooltip', () => {
     expect(onDismiss.calledOnce).toBe(true)
```
michalDob: assuming this change is ok will add a few specs for uncontrolled behavior before merging.

michalDob:
I am also considering adding handling for hover in addition to click into snacks since its a pretty common use case for tooltips.

Thoughts ?

PR #204, opened 2018-08-15
Pass snacksTheme to child components in certain form elements

wfro:
#196 allowed users to pass in a custom `snacksTheme` object. `Select`, `TextField`, and `PhoneNumberInput` all render `FloatingLabel` as a child component, which itself is wrapped with `withTheme`.

e.g. if a custom `snacksTheme` prop was provided to `TextField`, those theme values would be respected everywhere expect `FloatingLabel`, which was still reading it's values from `withTheme` (and by extension `themer`).

This commit passes the `snacksTheme` prop explicitly to `FloatingLabel` in those components so the theme colors are applied everywhere.

### DIFF
--- a/src/components/Forms/PhoneNumberField.js
+++ b/src/components/Forms/PhoneNumberField.js
@@ -263,6 +263,7 @@ class PhoneNumberField extends React.Component {
             hasError={hasError}
             htmlFor={inputId}
             style={{pointerEvents: 'none'}}
+            snacksTheme={snacksTheme}
           />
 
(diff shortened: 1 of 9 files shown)


wfro:
On the fence as to whether we needed more snapshots, but I ended up going for it. I tested these out locally and then committed the snapshots, the goal mainly being to test visual style.

dcocchia:
Ah, nice. I didn't even think about all the other components using floating label :D

PR #212, opened 2018-08-22
Add Pill component

Brantron:
This PR adds the `Pill` Component to snacks. The one question I had was around the color API. Currently it will throw a proptype warning if it is not a color in our styleguide which would ensure we are staying on brand while using this component. This may not be what we want moving forwards @mmarcuccio @dcocchia thoughts?

screen shot 2018-08-21 at 11 06 18 pm

### DIFF
--- a/src/components/Pill/Pill.js
+++ b/src/components/Pill/Pill.js
@@ -1,17 +1,15 @@
 import React from 'react'
-import { colors } from '../../styles'
 import { spacings } from '../../styles/spacing'
 import PropTypes from 'prop-types'
 import Radium from 'radium'
 import withTheme from '../../styles/themer/withTheme'
 
 const getStyles = props => {
-  const { color, hovorColor } = props
-  const restingColor = colors[color]
-  const onHoverColor = colors[hovorColor] || color
+  const { color: colorProp, snacksTheme } = props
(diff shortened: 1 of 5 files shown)


Brantron:
@spencerschack to be honest I was leaning towards making this more relaxed, but I figured the more stringent I made it the easier it would be to respond to feedback. I'm open to a string prop.

dcocchia:
I wonder if we could optionally pass a theme config attribute as the color? That way these could be theme-able or hard coded to a specific color.

Brantron:
@dcocchia @mmarcuccio thoughts?

screen shot 2018-08-23 at 2 41 47 pm

src/components/Pill/Pill.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { spacings } from '../../styles/spacing'
+import PropTypes from 'prop-types'
+import Radium from 'radium'
+import withTheme from '../../styles/themer/withTheme'
+
+const getStyles = props => {
+  const { color: colorProp, snacksTheme } = props
+  const isThemeColorKey = Object.keys(snacksTheme.colors).includes(colorProp)
```
spencerschack: `snacksTheme.colors[colorProp]` might be a simpler check

PR #216, opened 2018-08-27
export Pill component

Brantron:
This allows the Pill component to be used. To verify I built the package and pasted the dist in my node modules, the pill rendered without error.


Brantron:
That would certainly be a good idea. I might add that later

PR #222, opened 2018-08-28
position ToolTip component correctly, regardless of css translations

### LINKED ISSUE #224: Tooltips not correctly positioned when css transform translate is used
When a `Tooltip` component has a parent node with css transform translate applied, the resulting tooltip position is incorrect

Brantron:
The Tooltip component currently calculates styles incorrectly if there is a css transform translate applied in the node ancestry (think places like the checkout sidebar). This PR fixes that

Fixes #224 

screen recording 2018-08-28 at 04 58 pm

### DIFF
--- a/src/components/Tooltip/TooltipPosition.js
+++ b/src/components/Tooltip/TooltipPosition.js
@@ -46,18 +46,10 @@ class TooltipPosition extends PureComponent {
   getRect(node) {
     const container = node.getBoundingClientRect()
     const offset = {
-      top: 0,
-      left: 0,
+      top: container.top + window.pageYOffset,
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/docs/Tooltip.md
```diff
@@ -8,6 +8,23 @@ initialState = {
   style: 'primary',
   trigger: 'button',
   isVisible: false,
+  translate: 'none'
+}
+
+styles = {
```
dcocchia: The translate docs are cool for testing, but I don't think this is helpful to have in the docs for users or designers, etc. Can we remove them?

Brantron:
@dcocchia done!

PR #227, opened 2018-08-28
fixes #228 add no padding option to tooltip

### LINKED ISSUE #228: add no padding prop to tooltip
Currently the tooltip has padding which doesn't allow the developer to match the design specs if they differ- being able to pass in a `noPadding` prop would solve this.

scottjason:
fixes #228 

This enables us to use the tooltip without having pre-defined padding so we can match what we're building exactly w the design specs.

### DIFF
--- a/src/components/Tooltip/__tests__/Tooltip.spec.js
+++ b/src/components/Tooltip/__tests__/Tooltip.spec.js
@@ -6,6 +6,10 @@ import { mount }        from 'enzyme'
 import { spy, stub }    from 'sinon'
 import Tooltip          from '../Tooltip'
 
+console.log = s => {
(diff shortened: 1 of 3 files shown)


src/components/Tooltip/__tests__/Tooltip.spec.js
```diff
@@ -6,6 +6,10 @@ import { mount }        from 'enzyme'
 import { spy, stub }    from 'sinon'
 import Tooltip          from '../Tooltip'
 
+console.log = s => {
```
mmarcuccio: Remove console.log

scottjason:
Closing in favor of #233

PR #230, opened 2018-08-29
fixes #229 add option for drop shadow to tooltip

### LINKED ISSUE #229: add drop shadow option for tooltip
This is needed to match design specs and the tooltip is the outermost container- when used, the shadow should default to the same as what it is in the quick add component.

scottjason:
Fixes #229

cc: @mmarcuccio

### DIFF
--- a/src/components/Tooltip/InnerToolTip.js
+++ b/src/components/Tooltip/InnerToolTip.js
@@ -64,7 +64,8 @@ class InnerToolTip extends PureComponent {
       'medium',
       'large',
     ]),
-    arrowPosition: PropTypes.shape({})
+    arrowPosition: PropTypes.shape({}),
+    noPadding: PropTypes.bool,
   }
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/InnerToolTip.js
```diff
@@ -73,12 +74,15 @@ class InnerToolTip extends PureComponent {
   }
 
   get contentStyles() {
-    const { size, style, snacksStyle } = this.props
+    const { size, style, shadow, snacksStyle } = this.props
+    const resolvedStyle = shadow ?
+      {...style, ...{boxShadow: 'rgba(0, 0, 0, 0.25) 0px 3px 10px'}} :
+      {...style}
```
wfro: This can just be `style`

scottjason:
Closing in favor of #233

PR #231, opened 2018-08-29
Add typography

wfro:
### What
Add typography to snacks. This PR is open as an RFC - I'd love to get everyone's feedback on a few things around the API:

* What component name do we want? I chose `<Text />` to start, but `<Typography />` is also an option.
* What props do we want? Here's the React Material UI typography component for inspiration: https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Typography/Typography.js. On one hand, we could stick with a `style` prop which gives users more flexibility. Or, we could have more specific props like `gutterBottom`, `alignCenter`, etc. I like a `style` prop since it keeps the number of props down, but that might not be true for all.
* Generally open to ideas on how folks might use this, or specific use cases you already have. e.g. should we add specific color props so users can hook into snacks theme colors?
* What html tags should we map to each variant? Check the `htmlTapMapping` object for what I have so far.

### TODO
- [x] Finalize the API
- [x] More specs
- [x] Finish docs (need to add a table column for font size/weight/etc values to match the design system doc)

### DIFF
--- a/src/components/Typography/Text.js
+++ b/src/components/Typography/Text.js
@@ -17,48 +17,57 @@ const baseStyles = {
   color: colors.GRAY_13
 }
 
+// Matches the Open Sans font weights
+// https://fonts.google.com/specimen/Open+Sans?selection.family=Open+Sans
+const fontWeights = {
+  light: 300,
+  regular: 400,
+  semiBold: 600,
+  bold: 700
+}
+
 const variantStyles = {
   'T.92': {
     fontSize: '92px',
     lineHeight: '138px',
     letterSpacing: '0',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.82': {
     fontSize: '82px',
     lineHeight: '123px',
     letterSpacing: '0',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.72': {
     fontSize: '72px',
     lineHeight: '108px',
     letterSpacing: '0',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.64': {
     fontSize: '64px',
     lineHeight: '96px',
     letterSpacing: '0',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.58': {
     fontSize: '58px',
     lineHeight: '87px',
     letterSpacing: '0',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.46': {
     fontSize: '46px',
     lineHeight: '69px',
     letterSpacing: '0',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.36': {
     fontSize: '36px',
     lineHeight: '54px',
     letterSpacing: '0.03em',
-    fontWeight: 400
+    fontWeight: fontWeights.regular
   },
   'T.28': {
     fontSize: '28px',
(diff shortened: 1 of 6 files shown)


dcocchia:
The toughest part here are the titles' element types. Since a header level (h1, h2, etc) has semantic meaning within the context the header is being used, it's hard/impossible to guess which one will be correct. But using `h1` for everything also guarantees that we get it wrong and multiple `h1`s appear in the same hierarchy. 

Maybe we do a best guess mapping of the titles to header level?
title1: `h1`
title2: `h2`
title3: `h3`
sectionTitle: `h4`
bodyTitle: `h5`
bodyTitle2: `h6`

We still allow the override, but get a best guess at what hierarchical level makes sense?

wfro:
Here's a screenshot of the styleguide so far:
screen shot 2018-08-29 at 2 31 45 pm

Brantron:
can we allow style overrides? Something like

```js
  <Text variant="sectionTitle" overrideStyle="title1">
    Section title
  </Text>
```

So we can get max SEO flexibility out of the box?

dcocchia:
@Brantron - I think the point here is that these components control styling, rather than structure (which H-level us being used). This is why we allow overriding the `elementType` prop. So in your above scenario we would do:

  ```js
<Text
    variant="sectionTitle"
    elementType="h1"
>
    Section title
</Text>
```

**Comment by Brantron** [comment]
Ahh, I missed that. Awesome!

**Comment by mdboop** [comment]
When will this be merged? We want to use this in RPP! 😄 

Thanks @wfro!

**Comment by michalDob** [comment]
Could we get this in? I would love to clean up RPP Typography and switch it to snacks @wfro :)

**Comment by wfro** [comment]
Sorry all for the delay, I was finally able to see this one through!

The props API here is minimal so there's a decent chance we'll want to add a few helper props (e.g. gutters) once folks start using the component. That being said I'm fairly confident with the base API and am leaning towards merging as is, and will make any changes based on feedback.

File: src/components/Typography/Text.js [hunk 1]
```diff
@@ -0,0 +1,168 @@
...
+  'T.18': 'p',
+  'T.16': 'p',
+  'T.14': 'p',
+  'T.12': 'p',
+  'T.11': 'p'
+}
+
+// Matches the Open Sans font weights
+// https://fonts.google.com/specimen/Open+Sans?selection.family=Open+Sans
+const fontWeights = {
```
michalDob: should we be using these const in the font definitions ? It feels like the ones defined in font are just defaults out of this list 🤷‍♀️

src/components/Typography/Text.js
```diff
@@ -0,0 +1,168 @@
...
+  'T.11': {
+    fontSize: '11px',
+    lineHeight: '16px',
+    letterSpacing: '0.05em',
+    fontWeight: 400
+  }
+}
+
+const htmlTagMapping = {
+  'T.92': 'h1',
```
michalDob: I hate these name, but I am assuming they coming from snacks design system?

arguments against:
1. The correspond to font sizes they use, but font sizes change in design systems over time and I dont think we should have to refactor variants when the style changes.
2. Its easier to read the jsx if variants are somethnig like 'largeHeader', 'title' rather than T.58, T.28

I would much rather see names that suggest a use-case, even if they not perfect.

mdboop: Agreed. I would much rather see the components named `Header1` through `Header10` or however many we have. It's natural, and consistent with html tags, as well. You can easily extend that naming convention to `Body`, `Display`, or whatever kinds of text components we want.

wfro: Correct, the names do come from the design system. I'm mostly in agreement with your points, but that's more of a larger discussion to be had org-wide.

src/components/Typography/Text.js
```diff
@@ -0,0 +1,168 @@
...
+    fontSize: '36px',
+    lineHeight: '54px',
+    letterSpacing: '0.03em',
+    fontWeight: fontWeights.regular
+  },
+  'T.28': {
+    fontSize: '28px',
+    lineHeight: '42px',
+    letterSpacing: '0.05em',
+    fontWeight: 700
```
michalDob: fontWeights.bold

Issue #235, opened 2018-08-31
React element type comparison is broken by react-hot-loader


mmarcuccio:
Snacks components which depend on React element type comparison do not function correctly when used from an environment running react-hot-loader.  This is a [known issue with react-hot-loader](https://github.com/gaearon/react-hot-loader/issues/304) and not necessarily an issue in Snacks.  However, since react-hot-loader is used so widely, we should update snacks for now to patch this incompatibility.

Here is an example of snacks code that is affected by this bug:
https://github.com/instacart/Snacks/blob/ac492a63892d750284e6de4039fc52a225c2b123/src/components/Menus/Menu.js#L155

We should avoid importing the `areComponentsEqual` function from react-hot-loader since it creates an unnecessary runtime dependency.

For now let's create a utility function to get the runtime type of a given React component as suggested [here](https://github.com/gaearon/react-hot-loader/issues/304#issuecomment-223222772).  If this bug is ever fixed in react-hot-loader, we can refactor it out.  One potential concern is performance since this will instantiate an extra React element at each comparison.  We can memoize this function to reduce overhead.

PR #241, opened 2018-09-05
Fix typo to fully resolve RHL issue

mmarcuccio:
The first attempt at fixing the react-hot-loader issues on dev machines was unsuccessful. This was due to incorrect import syntax.


mmarcuccio:
This should fully fix #235

PR #249, opened 2018-09-11
refactoring/codeclimate

Brantron:
will be pushing commits

### DIFF
--- a/src/__tests__/isEqual.spec.js
+++ b/src/__tests__/isEqual.spec.js
@@ -10,32 +10,28 @@ describe('isEqual', () => {
     })
   })
   describe('with arrays', () => {
+    const first = [1,'s', {s: 'f'}]
     it('returns true when equal', () => {
-      const first = [1,'s', {s: 'f'}]
       const second = [1,'s', {s: 'f'}]
       expect(isEqual(first, second)).toBe(true)
     })
     it('returns false when not equal', () => {
-      const first = [1,'s', {s: 'f'}]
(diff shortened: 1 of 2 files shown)


Brantron:
https://codeclimate.com/github/instacart/Snacks/compare/brantron/refactor#ratings

Brantron:
I would say it was 50/50. I feel like a tool like this is great as an overall guide but I don't think striving for all A's should be the goal (not that anyone stated it is the goal). There were some improvements I made that felt good but to get rid of all of the issues I definately felt I was refactoring for the tool vs maintainability and readibility. Code golf is fun but I felt like I was code golfing vs trying to improve a shared lib. I think I would prefer to go back and reintroduce some offenders before merging this.

Brantron:
I will push a few more commits later. They will reduce the file scores but I'm curious where the happy medium falls. Maybe a B or greater is a good goal.

dcocchia:
Great feedback :) If you remove some of the code changes, let’s keep track of which rules seem arbitrary or too strict. We can change the code climate config to match our needs

Brantron:
Hmm, I reintroduced some of the code that was offending before and it's no longer an issue... I was definitely getting hits on too many return statements and too many args for a function. These are good rules overall and I think we would get value in keeping them.

dcocchia:
Interesting 🤔 Well, I think we can safely merge and keep track of our feelings on the code climate reports with future PRs to adjust how strict the rules are

PR #252, opened 2018-09-15
use longhand rules for spacing to avoid radium warnings

### LINKED ISSUE #246: Update the spacing module to only use longhand values
Using a mix of shorthand and longhand values creates warnings in Radium.  We should update spacing.* to only export longhand values.

Before:
```javascript
spacing.PADDING_SM  // => { padding: 16 }
```

After:
```javascript
spacing.PADDING_SM  // => { paddingTop: 16, paddingRight: 16, paddingBottom: 16, paddingLeft: 16 }
```

Brantron:
Fixes #246

### DIFF
--- a/src/styles/spacing/index.js
+++ b/src/styles/spacing/index.js
@@ -19,17 +19,33 @@ export const spacings = {
   HUGE: 64
 }
 
+const createXRules = (type, value) => ({
+  [`${type}Left`]: value,
+  [`${type}Right`]: value
+})
+
+const createYRules = (type, value) => ({
+  [`${type}Top`]: value,
+  [`${type}Bottom`]: value
+})
+
 const finalSpacings = {}
 Object.keys(spacings).forEach(spacing => {
   const pxValue = spacings[spacing]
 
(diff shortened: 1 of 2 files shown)


src/styles/spacing/index.js
```diff
@@ -19,17 +19,33 @@ export const spacings = {
...
 
+const createXRules = (type, value) => ({
+  [`${type}Left`]: value,
+  [`${type}Right`]: value
+})
+
+const createYRules = (type, value) => ({
+  [`${type}Top`]: value,
+  [`${type}Bottom`]: value
+})
```
dcocchia: Should we memoize these? It's possible they'll be called very frequently if someone is using them in a `render()`

mmarcuccio: I don't think we need to memoize these because it looks like they are used internally only to generate the spacing object.  That simple spacing object is then what could be called many times by the user.

dcocchia: ah yeah, that makes sense :)

PR #257, opened 2018-09-27
Allow for checkboxes that are both selected and disabled

michalDob:
Just to see if we can reach a consensus here, will update tests if we decide this is the way to go.

### DIFF
--- a/src/base/RadioCheckboxBase.js
+++ b/src/base/RadioCheckboxBase.js
@@ -40,7 +40,7 @@ const STYLE = {
 const getInputStyles = (props) => ({
   width: INPUT_BTN_SIZE,
   height: INPUT_BTN_SIZE,
-  fill: props.snacksTheme.colors.action
+  fill: props.isDisabled ? colors.GRAY_74 : props.snacksTheme.colors.action
 })
 
 function imgValidator (props, propName) {
@@ -87,14 +87,14 @@ class RadioCheckboxBase extends React.PureComponent {
   }
 
   state = {
-    isSelected  : this.props.isDisabled ? false : this.props.isSelected,
+    isSelected  : this.props.isSelected,
   }
 
   componentWillReceiveProps(nextProps) {
-    const { isDisabled, isSelected } = nextProps
+    const { isSelected } = nextProps
 
-    if (this.props.isSelected !== isSelected || this.props.isDisabled !== isDisabled) {
-      this.setState({isSelected: isDisabled ? false : isSelected})
+    if (this.props.isSelected !== isSelected ) {
+      this.setState({isSelected: isSelected})
     }
   }
 
@@ -113,8 +113,7 @@ class RadioCheckboxBase extends React.PureComponent {
     const { aria, bkgSvgSprites, btnType, isDisabled, id, style, value } = this.props
     const { isSelected } = this.state
 
-    if (isDisabled) { SvgComponent = bkgSvgSprites.disabled }
-    else { SvgComponent = isSelected ? bkgSvgSprites.selected : bkgSvgSprites.base }
+    { SvgComponent = isSelected ? bkgSvgSprites.selected : bkgSvgSprites.base }
 
     return (
       <div style={{...STYLE.button, ...style.button}}>
--- a/src/base/RadioCheckboxBase.js
+++ b/src/base/RadioCheckboxBase.js
@@ -37,10 +37,14 @@ const STYLE = {
   },
 }
 
+const getEnabledColor = (props) => {
+  return props.isSelected ? props.snacksTheme.colors.action : colors.GRAY_46
+}
+
 const getInputStyles = (props) => ({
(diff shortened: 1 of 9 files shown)


src/base/RadioCheckboxBase.js
```diff
@@ -113,8 +113,7 @@ class RadioCheckboxBase extends React.PureComponent {
     const { aria, bkgSvgSprites, btnType, isDisabled, id, style, value } = this.props
     const { isSelected } = this.state
 
-    if (isDisabled) { SvgComponent = bkgSvgSprites.disabled }
```
michalDob: will obviously remove support for disabled sprite if we go with this, its currently the same as base for both checkbox and radio sans color

src/components/Buttons/__tests__/Checkbox.spec.js
```diff
@@ -27,7 +27,7 @@ describe('Checkbox', () => {
     })
   })
 
-  it('sets isSelected to false if isDisabled is set to true', () => {
+  it('disables the input when it is disabled', () => {
```
michalDob: shouldn't do test refactoring at the same time I am changing code, but just for future reference:

How do you feel about these tests ... I am deeply uncomfortable with this. A lot of our snacks tests seem to be testing components through inspecting internal implementation by either examining properties of subcomponents or inspecting values of internal state and/or props. 

Is this the lay of the land in the react community nowadays? It feels super brittle to me, shouldn't tests be concerned with outside effects (such as colors of the component and whether or not it responds to click events in this example) and not in its internal guts.

wfro:
LGTM! Can you add a few screenshots of the visual changes?

src/components/Buttons/__tests__/Checkbox.spec.js
wfro: In general I'm 100% with you, I agree with avoiding testing internals (so I'd rarely/never set a prop directly or call setState from within a test). The only way we should be interacting with components in test is how a user would (e.g. click/change events).

I'm unclear on what you mean exactly with "examining  properties of subcomponents" though, I generally don't have a problem with assertions like `expect(htmlBtn.checked).toBe(true)` since it's checking actual DOM output and not component internals.

wfro:
This component is also used for `Radio`, so worth checking if that still functions the same way.

src/components/Buttons/__tests__/Checkbox.spec.js
michalDob: yeah maybe you are right, I was scared by some tests that where accessing states before.

Just feel like wrapper.find('input').instance() is a bit problematic because it relies on the input element actually being rendered inside a component ... but maybe I am paranoid.

michalDob:
radio got the same behavior change .. you can now disable redio but keep one of them selected which also makes sense to me. Screenshots on their way.

src/components/Buttons/__tests__/Checkbox.spec.js
michalDob: example form tooltip spec:
expect(tooltip.props().isVisible).toEqual(true)
    expect(tooltip.state().show).toEqual(false)

wfro: Yeah, agreed on the last two examples. The tests accomplish what they want but rely too much on implementation details. but it's similar to testing private methods in ruby, best to avoid it and if something is complex enough it can be extracted out and tested on its own. `wrapper.find('input').instance()` feels ok because it's finding the rendered output, I'm def open to arguments either way though.

michalDob: yeah fair enough, I think thats reasonable ... the find('input') is no different from snapshot test in a way.

lets just not do props and state inspection if we can avoid it :)

michalDob:
screen shot 2018-09-29 at 5 09 22 pm
screen shot 2018-09-29 at 5 10 10 pm
new behavior ^

michalDob:
screen shot 2018-09-29 at 5 43 40 pm
screen shot 2018-09-29 at 5 42 57 pm
old behavior ^

michalDob:
all the above checkboxes are in the 
selected
neither
selected and disabled
disabled
states respectively

src/base/RadioCheckboxBase.js
```diff
@@ -87,14 +90,14 @@ class RadioCheckboxBase extends React.PureComponent {
...
+    isSelected  : this.props.isSelected,
   }
 
   componentWillReceiveProps(nextProps) {
-    const { isDisabled, isSelected } = nextProps
+    const { isSelected } = nextProps
 
-    if (this.props.isSelected !== isSelected || this.props.isDisabled !== isDisabled) {
-      this.setState({isSelected: isDisabled ? false : isSelected})
+    if (this.props.isSelected !== isSelected ) {
```
wfro: nit - whitespace before closing paren

PR #263, opened 2018-10-11
MaskedTextField + DateField

apiv:
Adds a masked DateField component, and abstracts MaskedInputField from PhoneNumberField.


dcocchia:
@apiv This looks good. Looks like there are a number of tests failing, however. Can you take a look at those?

apiv:
I'll fix those now

PR #268, opened 2018-10-17
client-side a11y autofocus support for Form when fail to validate

TanakritBenz:
## What
- Add `onValidationError` callback to `Form`
- Add `triggerFocus` function to `TextField` and `PhoneNumberField`

## Why
Allow users of the library to handle client-side validation errors.
For example, one can use the callback to support a11y autofocus
```
onValidationError={(invalidComponents) => {
  if (invalidComponents && invalidComponents.length > 0) {
    invalidComponents[0].triggerFocus()
  }
}}
```

## Who should review this
@nbwar @dcocchia

### DIFF
--- a/src/components/Forms/Form.js
+++ b/src/components/Forms/Form.js
@@ -5,13 +5,15 @@ import isEqual from '../../utils/isEqual'
 class Form extends React.Component {
   static propTypes = {
     /** Form html chilren */
-    children    : PropTypes.node,
+    children         : PropTypes.node,
     /** HTML form attributes */
-    formProps   : PropTypes.shape({}),
+    formProps        : PropTypes.shape({}),
     /** onSubmit callback will pass in model as parameter */
-    onSubmit    : PropTypes.func,
+    onSubmit         : PropTypes.func,
+    /** onValidationError callback will pass in invalid FormComponents as parameter*/
+    onValidationError: PropTypes.func,
     /** errors from server mapped to model names. Will attach serverErrors styling to FormComppnents */
-    serverErrors: PropTypes.shape({})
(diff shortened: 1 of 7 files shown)


src/components/Forms/Form.js
```diff
@@ -67,7 +70,8 @@ class Form extends React.Component {
 
   formIsValid() {
     const components = Object.values(this.formComponents)
-    return components.every( component => component.validate() )
```
TanakritBenz: @nbwar The reason why I make the change here is because `Array.prototype.every()` immediately returns `false` if found. This can lead to a somewhat confusing UX like below ⬇️
screen shot 2018-10-16 at 10 42 48 pm
Last name is actually valid here but red because it failed on a prior submit. 
Let me know you think :)

src/components/Forms/FormComponent.js
```diff
@@ -40,6 +40,12 @@ const formComponent = (WrappedComponent) => {
       this.context.ICFormable && this.context.ICFormable.unregisterComponent(this)
     }
 
+    triggerFocus = () => {
```
nbwar: Don't like this function being defined within `FormComponent` as it really has nothing to do with the ICFormable Library

TanakritBenz: hmm, how should I do it then? I need a way to propagate up `triggerFocus()` from `TextField` and `PhoneNumberField` components to this wrapper. Without this, users would have to do `invalidComponent.FormComponent.triggerFocus()` which I feel kinda counter-intuitive. The way I see it, this is similar to `getValue()` and `hasValue()` that already in ICFormable.

PR #269, opened 2018-10-17
Can haz expanded border radius properties?

spencerschack:
Expand the `borderRadius` style property of `Button` into `borderRadiusTopLeft` etc. so users can override these properties to make things like segmented controls.


dcocchia:
I'm wondering more if we should add a lint rule to require expanded css properties 🤔

spencerschack:
@dcocchia +1 to that

spencerschack:
Oh man I would hate to define borders then:
```js
// Using 'border' with recursion returns an array of arrays
cssShorthandProps.expand('border', true);
/*
[
  ['border-top-width', 'border-right-width', 'border-bottom-width', 'border-left-width'],
  ['border-top-style', 'border-right-style', 'border-bottom-style', 'border-left-style'],
  ['border-top-color', 'border-right-color', 'border-bottom-color', 'border-left-color']
]
*/
```

PR #273, opened 2018-10-17
Update React dependency to 16.4.2

mmarcuccio:
This gives us support for forwarded refs.


mmarcuccio:
We should update the peer dependency version range when we start using any of the new features that are no backwards compatible (like forwarded refs).

Issue #126, opened 2018-03-14
Expose refs for all components


dcocchia:
Many of our components do not set a `ref` on their main output, so users of the library will be forced to use [findDOMNode](https://reactjs.org/docs/react-dom.html#finddomnode) to do things like focus on a button or input.

We should add a `ref` to each component to help with this.

tmarshall:
So can't be exposed using `ref` since the HoCs (`Radium()`, `withTheme`) don't forward.

We can add a method like `getInstance` to _every_ component to get around this.

But, ref forwarding [landed in 16.3](https://github.com/reactjs/rfcs/pull/30) - so we could choose to do 2 things; work around it with `getInstance` or put in support for 16.3 release and avoid adding a method that will later be removed

tmarshall:
^ though it seems like it'd be better if Radium did this

tmarshall:
I will build in a getter for now, and we can decide

dcocchia:
I think a nice way to do this would be to use the new ref forwarding api, use the same ref name for all components, and add tests to ensure we don't break that contract/api. It would require latest React, but maybe we could add in a fallback if the forwarding method isn't supported?

tmarshall:
@dcocchia sounds perfect. That's why I haven't started it, actually. Been hesitant to add in logic that may become moot once forwarded refs are in place.

mmarcuccio:
I am putting my work on hold while the Radium team is investigating.  If Radium ends up not supporting forward refs, we should discuss using the innerRef strategy that many other component libraries use.

mmarcuccio:
Forwardref is not going to be compatible with Radium. Let's move forward with the innerRef strategy.

PR #277, opened 2018-10-23
Add ref forwarding to Snacks components

mmarcuccio:
This is a WIP.  Adding ref forwarding to Snacks components.

Notes:
Enzyme does not currently [support forwarded refs](https://github.com/airbnb/enzyme/issues/1852). react-test-renderer does, but you must specify createNodeMock.

Radium does not support wrapping a component which is wrapped by React.forwardRef, so I am applying Radium HOC first before wrapping in React.forwardRef as a workaround.


mmarcuccio:
This is related to #126.

mmarcuccio:
This effort will be unblocked now due to switching to Emotion instead of Radium, however, it will be easier to start over due to the large amount of changes in master.

PR #278, opened 2018-10-29
Fix row style overrides

spencerschack:
The `Row` component styles were constructed under the assumption the style objects in the array passed as the `style` prop would be merged, however they are not. This change merges the style objects so that only one `grid` style gets applied.


mmarcuccio:
@spencerschack so Radium does not merge objects from the style array?

spencerschack:
It does not deep merge media queries: https://github.com/FormidableLabs/radium/blob/master/src/merge-styles.js#L33

PR #280, opened 2018-11-01
Add Switch component to snacks

michalDob:
Add a switch component to snacks, it has been in design for a while and we need it for some user accounts stuff in RPP


src/base/RadioCheckboxBase.js
```diff
@@ -35,16 +35,16 @@ const STYLE = {
   disabled: {
     color: colors.GRAY_74,
   },
-}
+})
 
-const getEnabledColor = (props) => {
-  return props.isSelected ? props.snacksTheme.colors.action : colors.GRAY_46
+const getEnabledColor = (props, state) => {
+  return state.isSelected ? props.snacksTheme.colors.action : colors.GRAY_46
```
michalDob: unrelated bug fix, color should be based on state not props

src/components/Buttons/docs/Checkbox.md
```diff
@@ -3,7 +3,7 @@ This component outputs a checkbox with an optional label.
 You can define whether the button starts in the pressed on unpressed state through the `isSelected` prop. In addition, you can change the selected/unselected state at any moment after the mount state by redifining the prop `isSelected`.
 
 ### Label
-Associating a label to every radio button is very important for accessibility purposes. There are two ways to do it. The easiest is to enclose the text with the component: `<Checkbox>My label</Checkbox>`. The other option is to associate a label yourself:
+Associating a label to every checkbox button is very important for accessibility purposes. There are two ways to do it. The easiest is to enclose the text with the component: `<Checkbox>My label</Checkbox>`. The other option is to associate a label yourself:
```
michalDob: opportunistic copy pasta fix

michalDob:
about code climate issue, do we have a pattern for sharing proptype shapes ?

dcocchia:
For accessibility, we need something other than the color to define if the switch is on or not. You can see how this looks on Instacart, for example:
screen shot 2018-11-01 at 5 28 29 pm
screen shot 2018-11-01 at 5 28 32 pm

https://www.instacart.com/store/account

Check out that component in the IC codebase for more info. One thing to note is that once the switch is enabled for high contrast, all of our switches on the site use the symbol to indicate on/off status. Try this out by enabling high contrast in account, then looking at https://www.instacart.com/store/account/notifications, for example.

michalDob:
Hmm interesting points, I will look at the one we use in customers, dont mind porting that one over if its close to snack ready.

as to accessibility: isn't the direction switch is pointing at enough 🤷‍♀️ 

as to animation: was on my agenda to add it at some point, just did not feel high priority and was more interested in having some kind of switch in snacks so I we dont have to touch call sites as we improve it

I am not a huge fan of red being off considering that our checkbox off is gray (and pretty much all the other switches I ever saw in other UI's use gray to mean off) 

Red screams to much as if off = bad to me but 🤷‍♂️ 

I noticed that the height of the switch in customers is 20px while the checkbox is 22px shouldn't they be the same ?

dcocchia:
"as to accessibility: isn't the direction switch is pointing at enough" - The issue is that users don't know which direction means off and which means on.

"I am not a huge fan of red being off considering that our checkbox off is gray (and pretty much all the other switches I ever saw in other UI's use gray to mean off)" - This feels like a design question. The switch used in the account page was made by Joran, so closest we have to an actual design system component. We could ask the current designers what they think?

The heights should probably be the same, yeah. I say go with what the snacks components are currently.

The animation can likely be copy/pasted from our current switch with some slight modification. :D

mmarcuccio:
There are really two different types of switches that should have different underlying DOM representations.

A on/off toggle that has one label:
[Switch] High contrast mode

Or a two-state toggle where each state has a unique label:
Easy [Switch] Hard

The first should be represented by a checkbox. The second should really be two radio buttons underneath where each label is associated with one radio button.  Maybe we should just start by supporting the first design and add a separate issue for also supporting the second?  I can definitely see someone wanting to use the second use case.

Also, one thing that could be useful is to add a bit of info to the switch documentation denoting when to typically use switch vs checkbox.  The biggest one that comes to mind for me is that switches should apply their "actions" immediately whereas checkboxes should require a separate "save" event.

src/components/Buttons/Switch.js
```diff
@@ -0,0 +1,43 @@
...
+const BKG_SVG_SPRITES = {
+  base:       bkgSvgBase,
+  selected:   bkgSvgSelected,
+}
+
+const Switch = (props) => {
+  return (
+    <RadioCheckboxBase
+      btnType='checkbox'
+      switch
```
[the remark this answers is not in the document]
michalDob: ohh my bad, will remove.

src/components/Buttons/Switch.js
```diff
@@ -0,0 +1,43 @@
...
+  style         : PropTypes.shape({
+    button        : PropTypes.object,
+    label         : PropTypes.object,
+    wrapEl        : PropTypes.object,
+  }),
+  value         : PropTypes.string,
+  wrapEl        : PropTypes.string,
+}
+
+export default Switch
```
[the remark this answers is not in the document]
michalDob: 👍

michalDob:
@dcocchia I discussed the off color with RPP designer and he agrees but will make sure other designers concur

michalDob:
New screenshot including accessibility:
screen shot 2018-12-18 at 10 29 34 pm

package.json
```diff
@@ -86,6 +86,7 @@
   },
   "jest": {
     "coverageDirectory": "./coverage/",
+    "testURL": "http://localhost",
```
michalDob: without specifying this specs blow up for me locally with: SecurityError: localStorage is not available for opaque origins

src/components/Buttons/Switch.js
```diff
@@ -0,0 +1,97 @@
...
+  name          : PropTypes.string,
+  onChange      : PropTypes.func,
+  style         : PropTypes.shape({
+    button        : PropTypes.object,
+    label         : PropTypes.object,
+    wrapEl        : PropTypes.object,
+  }),
+  value         : PropTypes.string,
+  wrapEl        : PropTypes.string,
+  isAccessible    : PropTypes.bool
```
michalDob: Anyone can think of a better name for this prop, if so I am all ears.

dcocchia: We call this prop `enableOnOffLabel` in the store app.

michalDob: 👍 will change to it unless others have other ideas

michalDob: renamed to displayOnOffLabel

mmarcuccio: Maybe we should default this to true? That way we are accessible by default.

michalDob: I think this is usually enabled only if you enable high contrast no ? (at least most companies do it this way) 🤷‍♀️ But ours dont look to bad so I am fine either way, (really a design call I think)

PR #284, opened 2018-11-28
Add the ability to add attributes to inner elements

stefceror:
For the `NavigationPill` component, proper accessibility requires that certain attributes (eg `aria-selected` and `role=tab`) be placed on the inner `<a>` tag rather than on the `<li>` tag.  

Likewise, for the `NavigationPills` component certain elements should be placed on the inner `<ul>` tag rather than the `<div>` element

This change adds the ability to pass an `inner` object as part of the `elementAttributes` prop.  The change is fully backwards compatible and requires no changes to existing code.

### DIFF
--- a/src/components/NavigationPills/NavigationPill.js
+++ b/src/components/NavigationPills/NavigationPill.js
@@ -44,7 +44,6 @@ const styles = {
 const NavigationPill = props => {
   const { isActive, snacksTheme, text } = props
   const { primaryForeground } = snacksTheme.colors
-  const { inner, ...outer } = props.elementAttributes
   const activeStyles = {
     backgroundColor: primaryForeground,
     color: colors.WHITE,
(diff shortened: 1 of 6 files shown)


mmarcuccio:
Instead of using a generic 'outer' and 'inner' object, we could use specific named props where appropriate and apply the 'rest' props to the wrapper element by default.  For example:
Using elementAttributes:
```jsx
<NavigationPill elementAttributes={outer: {'aria-something'='foo' }, inner: {role: 'tab' }} />
```
Suggestion:
```jsx
<NavigationPill aria-something='foo' listItemProps={{role: 'tab'}} />
```

This means if we have a complex component where you need to apply custom attributes to more than one child component, we can support multiple named props.  It also reduces prop nesting.

stefceror:
The reason I chose to do it that way was mostly to maintain compatibility with existing components.  I like the idea of adding a specific group of props for a particular element, so maybe we could do a hybrid approach:

Mine:
```jsx (no need to specify outer explicitly)
<NavigationPill elementAttributes={{'aria-something':'foo', inner: {role: 'tab' }}} />
```

Your suggestion:
```jsx
<NavigationPill aria-something='foo' listItemProps={{role: 'tab'}} />
```

Hybrid:
```jsx
<NavigationPill elementAttributes={{'aria-something':'foo'}} listItemProps={{role: 'tab'}} />
```

PR #309, opened 2019-02-05
Add button role to select trigger

wfro:
This results in a few a11y improvements:
* Screen readers will better announce the menu button as actionable
* Screen reader focus is correctly returned back to the menu button -- Select was
  already returning focus to the menu button programatically, but
  because the menu button only had tabindex it wasn't seen by screen
  readers as an interactive element. Screen reader focus would be stuck
  on the selected menu item instead of being returned to the button.


wfro:
fyi @nbwar @dcocchia

PR #314, opened 2019-04-05
Allow native focus indicators on interactive elements

loganmurdock:
Fixes for accessibility purposes: https://instacart.atlassian.net/browse/CS-4922

Add focus indicator to textfields and buttons:

snacks_updates


src/components/Buttons/__tests__/__snapshots__/Button.spec.js.snap
```diff
@@ -561,9 +561,7 @@ exports[`Button renders correctly when disabled 1`] = `
   <button
     data-radium={true}
     disabled={true}
-    onBlur={[Function]}
     onClick={[Function]}
-    onFocus={[Function]}
```
dcocchia: 🤔 Interesting. Seems like Radium was somehow adding these to the output by having a `:focus` sudo style.

loganmurdock: Yeah, I was surprised that removing
```
':focus': {	
    outline: 'none'	
  }
```
had this impact on the snap shot.

PR #317, opened 2019-04-13
Support Overriding Default z-index Values for DropdownMenu and Tooltip

mmarcuccio:
Currently these two components use a hardcoded value for z-index.  Sometimes when using these components you need to adjust these values.

There were a couple of options for adding this support, but I think that the way with the best backwards compatibility is to add an optional z-index prop to override the existing values.

Another possible way is to expose style overrides for the parts of these components that use z-index, but I feel like that is less friendly to the user since they must understand implementation details to adjust the z-index.

### DIFF
--- a/src/components/Menus/DropdownMenu.js
+++ b/src/components/Menus/DropdownMenu.js
@@ -57,7 +57,9 @@ class DropdownMenu extends React.Component {
     /** Customize style of menu parent */
     menuContainerStyle: PropTypes.shape({}),
     /** Props passed to Menu component */
-    menuProps: PropTypes.shape({})
+    menuProps: PropTypes.shape({}),
+    /** Override the default z-index of the open DropdownMenu container */
+    zIndex: PropTypes.number,
   }
 
   static defaultProps = {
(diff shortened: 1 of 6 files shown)


mmarcuccio:
Waiting for Snacks PR resolution before continuing.

NinjaBanjo:
> We've already seen a need for a way to extensibly override styles for "edge cases"
> 
> I would be a fan of this being a styleOverride object, with a zIndex property on it, so when future use-cases come along we don't have an explosion of css properties as normal proptype (and inconsistent support for different props) across components
> 
> I see it looking very close to what you have here
> 
> ```
> <DropdownMenu styleOverrides={{ zIndex: 500 }}
> ```

further discussion in slack, there is an existing pattern of a `shape` prop with sub-properties for each sub-component in the snack component. that take a regular react-like style object.

This leads us to the following:
If we have a pattern (but don’t like it) we should introduce a new pattern or keep the existing and plan to spend time introducing one in the future.

It would be harder to convert everything if a bunch of one-off solutions pop up everywhere when we do find the good final solution

mmarcuccio:
Yep. Let's have another discussion on how best to handle the issue of style overrides in Snacks.  For now I will update this PR to use the existing system.

mmarcuccio:
After further review, it looks like this would be a breaking change. Tooltip and Dropdown do not currently use a nested style override approach. They both handle it differently and incompatibly with the nested style object approach.

I am going to put this PR on hold for now.  We need to have a discussion and agree on a standardized approach for this.

PR #318, opened 2019-04-18
Updating TextHint accessibility

ychoi1:
### Context (Links to Quips/Jira)
https://instacart.atlassian.net/browse/CS-7397
https://instacart.atlassian.net/browse/CS-7398

### Problem
Text hints for `MaskedTextField` are not accessibility compliant:
- Text hint was able to receive focus when not visible on screen
- Text hint was not associated to relative field
- Validation error was not associated to relative field

### Solution
- Added `visible: hidden / visible` css to TextFieldHint
- Added `aria-describedby` to `MaskedTextField`. Passing hint and validation error input id

### DIFF
--- a/src/components/Forms/MaskedTextField.js
+++ b/src/components/Forms/MaskedTextField.js
@@ -276,7 +276,7 @@ class MaskedTextField extends React.Component {
           />
 
           <TextFieldHint
-            inputId={inputId}
+            inputId={`hint_${inputId}`}
             text={maskHint}
             show={!hasValue && isFocused}
             disabled={disabled}
--- a/src/components/Forms/MaskedTextField.js
+++ b/src/components/Forms/MaskedTextField.js
@@ -275,7 +275,12 @@ class MaskedTextField extends React.Component {
             snacksTheme={snacksTheme}
           />
 
-          <TextFieldHint text={maskHint} show={!hasValue && isFocused} disabled={disabled} />
+          <TextFieldHint
+            inputId={inputId}
+            text={maskHint}
+            show={!hasValue && isFocused}
+            disabled={disabled}
+          />
 
           <MaskedTextInput
             mask={mask}
@@ -285,6 +290,7 @@ class MaskedTextField extends React.Component {
             name={name}
             aria-required={required}
             aria-invalid={hasError}
+            aria-describedby={hasError ? `hint_${inputId} error_${inputId}` : `hint_${inputId}`}
             onBlur={this.handleInputBlur}
             onChange={this.handleInputChange}
             onFocus={this.handleInputFocus}
--- a/src/components/Forms/TextField.js
+++ b/src/components/Forms/TextField.js
@@ -214,14 +214,10 @@ class TextField extends React.Component {
     const snacksStyles = getSnackStyles(snacksTheme)
 
     const inputId = id
+    const errorId = `error_${inputId}`
+    const hintId = `hint_${inputId}`
     const showHintText = hintText && !hasValue && isFocused
-    let describedByIds = ''
-    if (hasError) {
-      describedByIds += `error_${inputId}`
-    }
-    if (hintText) {
-      describedByIds += ` hint_${inputId}`
-    }
+    const describedByIds = `${hasError ? errorId : ''} ${hintText ? hintId : ''}`.trim()
 
     return (
       <div
@@ -266,7 +262,7 @@ class TextField extends React.Component {
             type={type}
             aria-required={required}
             aria-invalid={hasError}
-            aria-describedby={describedByIds ? describedByIds.trim() : null}
+            aria-describedby={hasError || hintText ? describedByIds : null}
             style={[
               styles.input,
               inputStyle,
--- a/src/components/Forms/TextField.js
+++ b/src/components/Forms/TextField.js
@@ -259,7 +259,12 @@ class TextField extends React.Component {
             type={type}
             aria-required={required}
             aria-invalid={hasError}
-            aria-describedby={[hasError ? `error_${inputId}` : null, hintText ? `hint_${inputId}` : null].join(' ').trim()}
+            aria-describedby={[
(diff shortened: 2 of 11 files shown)


src/components/Forms/MaskedTextField.js
```diff
@@ -285,6 +290,7 @@ class MaskedTextField extends React.Component {
             name={name}
             aria-required={required}
             aria-invalid={hasError}
+            aria-describedby={hasError ? `hint_${inputId} error_${inputId}` : `hint_${inputId}`}
```
wfro: Does a screenreader read the text from both of these elements if there's an error?

src/components/Forms/TextFieldHint.js
```diff
@@ -15,9 +15,11 @@ const styles = {
     transition: 'opacity 250ms cubic-bezier(0.23, 1, 0.32, 1) 0ms',
     pointerEvents: 'none',
     zIndex: 1,
+    visibility: 'hidden',
```
wfro: Just to double check here, is this specifically so that screen readers won't read this content?

src/components/Forms/MaskedTextField.js
```diff
@@ -275,7 +275,12 @@ class MaskedTextField extends React.Component {
             snacksTheme={snacksTheme}
           />
```
wfro: Is `MaskedTextField` the only text field that needs this? There's another `TextField` component, not sure if that's relevant here.

src/components/Forms/MaskedTextField.js
loganmurdock: Actually just learned about this on Monday at workshop. Yeah, aria-labelledby and aria-describedby accept multiple ids just like class names. Pretty cool.

ychoi1: Correct, it'll read off in order of each element listed

src/components/Forms/TextFieldHint.js
ychoi1: Right. When you tab around in the screen reader and the hint isn't showing, the hint will still receive focus by the screen reader and read off

src/components/Forms/MaskedTextField.js
wfro: Which element gets the `error_<input id>` class? I can't find one at a glance but I remember it possibly being part of the form system?

src/components/Forms/docs/DateField.md
```diff
@@ -0,0 +1,14 @@
+DateField Example:
```
wfro: Is this related to the other accessibility stuff?

src/components/Forms/TextFieldHint.js
```diff
@@ -35,17 +37,22 @@ class TextFieldHint extends Component {
...
   }
 
   render() {
-    const { disabled, show, style, text } = this.props
+    const { disabled, show, style, text, inputId } = this.props
 
     return (
-      <div style={[styles.root, style, disabled && styles.disabled, show && styles.show]}>
+      <div
+        id={`hint_${inputId}`}
```
wfro: more of a nit - since you have to define the hint_<x> class in the parent component anyways, you can pass in the full `hint_${inputId}` string instead of just `inputId`.

src/components/Forms/MaskedTextField.js
ychoi1: `ValidationError`

src/components/Forms/TextFieldHint.js
mmarcuccio: Also, this will interrupt the CSS transition since it happens immediately. You want to use multiple transitions, one on visibility and one on opacity.  The visibility transition should have a delay value equal to the duration of the opacity transition.

src/components/Forms/docs/DateField.md
ychoi1: I just added this since I think the Rx team added it, and didn't see an example listed for it.

src/components/Forms/MaskedTextField.js
ychoi1: Good call, didn't see TextField using hint. Will update if needed

src/components/Forms/TextFieldHint.js
mmarcuccio: It might be better for future reusability to apply the id prop directly here, and pass in `hint_${inputId}` from the parent component.

src/components/Forms/TextFieldHint.js
ychoi1: Wouldn't that be ok since the opacity will still be transitioning while visibility is already set? From testing, it looks like opacity transition still takes effect regardless of visibility.

mmarcuccio: Before it would both fade in and fade out. After this change it will only fade in. Setting visible false happens immediately.

src/components/Forms/TextField.js
```diff
@@ -252,7 +259,7 @@ class TextField extends React.Component {
             type={type}
             aria-required={required}
             aria-invalid={hasError}
-            aria-describedby={hasError ? `error_${inputId}` : null}
+            aria-describedby={[hasError ? `error_${inputId}` : null, hintText ? `hint_${inputId}` : null].join(' ').trim()}
```
ychoi1: Been having trouble getting this sort of logic to either be cleaner or follow complexity rules. Open to suggestions on refactoring this bit

mmarcuccio: Not that much better, but here's another approach:
```
aria-describedby={[hasError && `error_${inputId}`, hintText && `hint_${inputId}`].filter(Boolean).join(' ')
```

ychoi1:
Updated with review changes.

Note: `Select` also needs accessibility changes. Will create separate branch for it since the changes aren't minimal.

ychoi1:
Before merging, should we consider addressing the codeclimate issue?

PR #321, opened 2019-04-23
Add propTypes to DateField

apiv:
The propTypes of DateField are a subset of the propTypes of MaskedTextField. This PR adds the proptypes so they appear in the styleguide for DateField.

### DIFF
--- a/src/components/Forms/DateField.js
+++ b/src/components/Forms/DateField.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe'
-import MaskedTextField from './MaskedTextField'
+import MaskedTextField, { propTypes as maskedTextFieldPropTypes } from './MaskedTextField'
+import { omit } from 'underscore'
 
 const mask = [/\d/, /\d/, '/', /\d/, /\d/, '/', /\d/, /\d/, /\d/, /\d/]
 const hint = 'MM/DD/YYYY'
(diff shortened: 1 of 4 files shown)


src/components/Forms/DateField.js
```diff
@@ -1,15 +1,26 @@
 import React from 'react'
 import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe'
-import MaskedTextField from './MaskedTextField'
+import { omit } from 'underscore'
```
wfro: I'm not sure if we use underscore, I don't think we do?

wfro: Take it back, we still have it as a dependency but it's not being used anywhere anymore. If it's not too much of a pain can you do this without `omit`? Would help keep the size down.

src/components/Forms/MaskedTextField.js
```diff
@@ -91,78 +91,80 @@ const getInputSyles = ({ props, theme, isFocused }) => {
   }
 }
 
+export const propTypes = {
```
nathanmarks: Personally I would give this a more specific name since it's being exported. It's a more scalable approach taking IDE autocomplete features into consideration.

apiv: Went with a more specific naming convention.

src/components/Forms/DateField.js
apiv: Added a simple `omit` function to the `utils` folder instead.

PR #323, opened 2019-04-29
Updated docs

tmarshall:
I noticed that the docs for the responsive styles didn't show how to actually import `responsive` from snacks, and then I went down a bit of a rabbit hole.

I ended up upgrading Styleguidist and refactoring the readmes to have consistent formatting.


styleguide.config.js
```diff
@@ -5,7 +5,6 @@ module.exports = {
     var componentName = parts[parts.length - 1];
     return componentPath.replace(`/${componentName}`, `/docs/${componentName.replace('.js', '')}.md`);
   },
-  highlightTheme: 'pastel-on-dark',
```
tmarshall: This config attribute is no longer supported by styleguidist.

Looks like we'll need to do more granular styling if we want to.

PR #330, opened 2019-06-05
Removing aria-haspopup from Dropdownmenu.js

mdprescott:
Received a report from accessibility auditors that aria-haspopup is not well supported and is really meant for menu and submenus. Support for aria-haspopup is coming slowly, but in its current state adding an aria-haspopup to custom select drop downs doesn't do much. When the attribute becomes more fleshed out, it may be that having it on this element may cause unforeseen future issues. For now, it was recommend we remove this attribute.

See: https://hub.accessible360.com/kb/articles/151

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,11 +27,12 @@ jobs:
       - run:
           name: Run tests
           command: yarn test
-      
-      - run: 
+
+      - run:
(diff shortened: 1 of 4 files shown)


mdprescott:
This has been punted, will re-open if needed

PR #334, opened 2019-06-14
Typings

nathanmarks:
Adds typings.

For some reason, I had linting failing on a few specs because they weren't named `.spec.js` like the other tests... I renamed them `.spec.js` so had to rewrite the snapshots for them as well.


test/typescript/FooRefTest.tsx
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable */
```
mmarcuccio: Do we still need these files in this test directory?

nathanmarks: I was using these to test some typings -- i think it's useful to have them. But I can kill them if you want.

src/components/Forms/Form.d.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+type FormComponent = React.Component<FormComponentInjectedProps>
+
+export interface FormProps {
+  /** Form html chilren */
+  children?: React.ReactNode
+
+  /** HTML form attributes */
+  formProps?: React.ComponentPropsWithoutRef<'form'>
```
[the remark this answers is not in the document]
nathanmarks: Good catch!!!!

test/typescript/FooRefTest.tsx
mmarcuccio: I'm fine either way.

.eslintrc.js
```diff
@@ -27,6 +27,11 @@ module.exports = {
     'react/forbid-prop-types': [WARN, { forbid: 'any' }],
     'react/jsx-filename-extension': OFF,
     'react/jsx-no-literals': OFF,
+
+    'import/no-extraneous-dependencies': [
+      'error',
+      { devDependencies: ['scripts/**/*.js', '**/*.spec.js'] },
```
mmarcuccio: We ended up doing this in customers too.

Issue #333, opened 2019-06-12
Tooltip size small, horizontally placed, TooltipArrow overlaps inner text


KsenijaGogic:
When generating a small, horizontally positioned tooltip, i.e.,
```jsx
<Tooltip size="small" placement="left || right" />
```
the `<TooltipArrow />` arrow overlaps the inner content.

| `<Tooltip size="small" placement="left" />` | `<Tooltip size="small" placement="right" />`|
|--|--|
| <img width="271" alt="Screen Shot 2019-06-12 at 3 07 37 PM" src="https://user-images.githubusercontent.com/10468542/59379554-eb728200-8d24-11e9-8ee0-c622a01a2a33.png"> | <img width="287" alt="Screen Shot 2019-06-12 at 3 16 36 PM" src="https://user-images.githubusercontent.com/10468542/59379651-2c6a9680-8d25-11e9-9f78-a7e34446a15f.png"> |

KsenijaGogic:
I already added a fix for this locally, just waiting for GH permissions to push branch up ✌️

KsenijaGogic:
Solved w/ #335.

PR #335, opened 2019-06-17
[ISSUE-333] Fixes <Tooltip /> whitespace for <Tooltip size="small" />

KsenijaGogic:
## Context
https://github.com/instacart/Snacks/issues/333

## Problem
The left- and right-padding of small Tooltips was set to 0. This was a bit awkward when the tooltip was positioned `top` and `bottom` of target, but was particularly a problem for when the tooltip was positioned `right` or `left`, as the Tooltip Arrow would cover the inner content.

## Solution
Two small tweaks were made:
1. ~Wrapped inner content of tooltip to be `position: relative`, so that we can always ensure a higher z-index than non-crucial content.~
2. Turns out that the rouge issue here was of a silently failing import `...spacing.PADDDING_X_XS,` vs. `...spacing.PADDING_X_XS`.

## Screenshots

| `<Tooltip size="small" placement="left" />` | `<Tooltip size="small" placement="right" />`|
|--|--|
| Screen Shot 2019-06-12 at 3 07 37 PM | Screen Shot 2019-06-12 at 3 16 36 PM |
| Screen Shot 2019-06-17 at 1 40 53 PM | Screen Shot 2019-06-17 at 1 40 59 PM |

### DIFF
--- a/src/components/Tooltip/InnerToolTip.js
+++ b/src/components/Tooltip/InnerToolTip.js
@@ -50,9 +50,9 @@ const RESOLVED_COLOR = {
 const RESOLVED_SIZE = {
   small: {
     fontSize: '14px',
-    ...spacing.PADDDING_X_XS,
     paddingTop: '9px',
     paddingBottom: '9px',
+    ...spacing.PADDING_X_XS,
   },
   medium: {
     fontSize: '16px',
--- a/src/components/Tooltip/InnerToolTip.js
+++ b/src/components/Tooltip/InnerToolTip.js
@@ -23,6 +23,12 @@ const styles = {
   },
 }
 
+const spanStyles = {
+  root: {
+    position: 'relative',
+  },
+}
+
 const RESOLVED_COLOR = {
   primary: {
     background: colors.GREEN_500,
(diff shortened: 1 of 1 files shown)


nathanmarks:
@mmarcuccio would like to fix this -- will it screw up anything in customers that is compensating for it?

mmarcuccio:
> @mmarcuccio would like to fix this -- will it screw up anything in customers that is compensating for it?

Probably. We will need to find usages of the Tooltip component and check.

nathanmarks:
> > @mmarcuccio would like to fix this -- will it screw up anything in customers that is compensating for it?
> 
> Probably. We will need to find usages of the Tooltip component and check.

Could only find us passing this prop in one place: https://github.com/instacart/carrot/blob/master/customers/instacart/app/assets/javascripts/store_app/components/module_renderer/modules/Checkout/TotalsLineItem.jsx#L230

mmarcuccio:
> > > @mmarcuccio would like to fix this -- will it screw up anything in customers that is compensating for it?
> > 
> > 
> > Probably. We will need to find usages of the Tooltip component and check.
> 
> Could only find us passing this prop in one place: https://github.com/instacart/carrot/blob/master/customers/instacart/app/assets/javascripts/store_app/components/module_renderer/modules/Checkout/TotalsLineItem.jsx#L230

Awesome. It's definitely the most important to fix this here in Snacks.

PR #344, opened 2019-06-26
Add SVG component type declaration generation and verification

nathanmarks:
The generated SVG components weren't getting declarations generated. Didn't see this in a consuming project initially as the TS compiler allowed the global declaration for `.svg` modules in said project to apply to the imports ending in `.svg`.

This now generates declarations for those files.

I've also added two additional steps to the build verification process:

- Will verify that the TS compiler is satisfied with the declarations output into the `dist` dir
- Ensures that the ESM build bundles without error and performs the same checks against the resulting bundle as we perform against the UMD build.

### DIFF
--- a/scripts/release/build_esm.js
+++ b/scripts/release/build_esm.js
@@ -1,22 +1,6 @@
-import path from 'path'
-import fs from 'fs'
 import { execSync } from 'child_process'
 
 const exec = command => execSync(command, { stdio: 'inherit' })
-
-const svgrTempDir = path.resolve(process.cwd(), 'tmp/svgr')
-const svgrAssetsDir = path.resolve(svgrTempDir, 'assets')
-const svgrComponentsDir = path.resolve(svgrTempDir, 'components')
-
-function ensureDir(dir) {
-  if (!fs.existsSync(dir)) {
-    fs.mkdirSync(dir)
-  }
-}
-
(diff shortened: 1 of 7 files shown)


scripts/release/build_esm.js
```diff
@@ -1,32 +1,23 @@
...
-const svgrAssetsDir = path.resolve(svgrTempDir, 'assets')
-const svgrComponentsDir = path.resolve(svgrTempDir, 'components')
-
-function ensureDir(dir) {
-  if (!fs.existsSync(dir)) {
-    fs.mkdirSync(dir)
-  }
-}
-
-// svg will complain if the --out-dir does not exist
```
nathanmarks: apparently that doesn't seem to be the case... not sure why I encountered it originally

scripts/svgDtsTemplate.js
```diff
@@ -0,0 +1,16 @@
+function svgDtsTemplate({ types: t, template }, _, { componentName }) {
+  const typeScriptTpl = template.smart({ plugins: ['typescript'] })
+  const {
+    typeAnnotation,
+  } = typeScriptTpl.ast`type Foo = React.FunctionComponent<React.SVGProps<SVGSVGElement>>`
```
nathanmarks: Just doing this so i can easily grab the AST node for the type annotation and add it to the component identifier

PR #347, opened 2019-07-04
Adds props.customElement to <Button />

KsenijaGogic:
## Problem
Trying to figure out an elegant way to:
1. Allow Snacks users to pass a custom component to the `<Button />` component to leverage styling (perfect use case: `react-router`'s `<Link />`)
2. while maintaining the source / approach of `<Button />` mostly untouched

*The motivation for this is mainly to avoid improper HTML practices with nesting a `<Button />` inside a `<Link />`.*

## Solution
Current approach leverages existing dynamic element creation to add a 3rd option (`a`, `button`, `props.customElement`). ~Adds an additional `customElementProps` prop to allow users to specify props they may need (`to`, `q`, etc.).~ `customElementProps` removed in order to favour `elementAttributes`.

### DIFF
--- a/src/components/Buttons/Button.d.ts
+++ b/src/components/Buttons/Button.d.ts
@@ -40,11 +40,7 @@ export interface ButtonProps
    */
   href?: string
 
-  /**
-    * If passed, users will have the ability to render a custom element
-    * or React Component in the DOM while leveraging the styling of a
-    * Snacks Button (e.g., using react-router's Link)
-    */
+  /** A custom element to render instead of a button. */
   elementType?: React.ElementType
 }
 
--- a/src/components/Buttons/Button.d.ts
+++ b/src/components/Buttons/Button.d.ts
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { RadiumStyles, ElementAttributes } from '../..'
 import { WithThemeInjectedProps, ApplyWithTheme } from '../../styles/themer/withTheme'
 
+
 export interface ButtonProps
   extends WithThemeInjectedProps,
     Pick<
@@ -39,6 +40,13 @@ export interface ButtonProps
    * tag with the supplied href value.
    */
   href?: string
+
+  /**
+    * If passed, users will have the ability to render a custom element
+    * or React Component in the DOM while leveraging the styling of a
+    * Snacks Button (e.g., using react-router's Link)
+    */
+  elementType?: keyof JSX.IntrinsicElements | Function
 }
 
 declare const Button: ApplyWithTheme<React.ComponentClass<ButtonProps>>
--- a/src/components/Buttons/Button.d.ts
+++ b/src/components/Buttons/Button.d.ts
@@ -2,7 +2,6 @@ import * as React from 'react'
 import { RadiumStyles, ElementAttributes } from '../..'
 import { WithThemeInjectedProps, ApplyWithTheme } from '../../styles/themer/withTheme'
 
-
 export interface ButtonProps
   extends WithThemeInjectedProps,
     Pick<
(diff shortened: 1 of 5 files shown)


src/components/Buttons/Button.js
```diff
@@ -238,10 +238,14 @@ const Button = props => {
       props.onMouseDown(e, props)
     },
     ...props.elementAttributes,
+    ...props.customElementProps,
```
nathanmarks: do we need this as well as `elementAttributes`?

src/components/Buttons/Button.js
```diff
@@ -238,10 +238,14 @@ const Button = props => {
...
   }
   if (props.href) {
     finalProps.href = props.href
   }
+  if (props.customElement) {
```
nathanmarks: this is very opinionated and incorrect for some use cases

for eg: `<a href="xxx" role="button" />` doesn't make sense

i think we should remove this and let users pass it as necessary in `ElementAttributes`.

src/components/Buttons/Button.js
```diff
@@ -207,7 +207,7 @@ const getSnacksStyles = props => {
 const Button = props => {
   const snacksStyles = getSnacksStyles(props)
 
-  const ElementType = props.href ? 'a' : 'button'
+  const ElementType = props.customElement ? props.customElement : props.href ? 'a' : 'button'
```
nathanmarks: we should call the props `elementType` to be consistent with other snacks component APIs

src/components/Buttons/docs/Button.md
```diff
@@ -233,3 +233,23 @@ import { Button, Icon } from 'ic-snacks'
...
+import { Button, Icon, Link } from 'ic-snacks';
+
+  <div style={{padding: '24px'}}>
+    <span style={{marginRight: '24px'}}>
+      <Button snacksStyle="primary" size="small" customElement='span'>
+        Using a Custom HTML Element
+      </Button>
+    </span>
+    <span style={{marginRight: '24px'}}>
+      <Button snacksStyle="primary" size="small" customElement={Link} customElementProps={{ href: '/foo' }}>
```
nathanmarks: This isn't an example we should use since `Button` supports this use case already via the `href` prop.

```js
<Button href="/foo">Foo</Button>
```

src/components/Buttons/Button.js
```diff
@@ -319,6 +323,15 @@ Button.propTypes = {
 
   /** Snacks theme attributes provided by `Themer` */
   snacksTheme: themePropTypes,
+
+  /**
+    If passed, users will have the ability to render a custom element
+    or React Component in the DOM while leveraging the styling of a
+    Snacks Button (e.g., using react-router's Link)
+  */
+  customElement: PropTypes.node,
```
nathanmarks: This proptype isn't correct.. this would be `element={<Component />}` rather than `element={Component}`

src/components/Buttons/Button.js
KsenijaGogic: I think intentionally separating `attributes` and `props` is valuable (from a taxonomical perspective.) What do you think? :)

src/components/Buttons/Button.d.ts
```diff
@@ -39,6 +39,13 @@ export interface ButtonProps
    * tag with the supplied href value.
    */
   href?: string
+
+  /**
+    * If passed, users will have the ability to render a custom element
```
nathanmarks: We should just describe the prop, not say what happens the user passes it.

Plus, if they pass it, they don't have the ability to decide,

> A custom element to render instead of a `button`. Will receive all props usually passed to the button element.

or something like that

KsenijaGogic:
Closing this PR as a [new release of React Router will solve this problem specifically].(https://github.com/ReactTraining/react-router/pull/6462)

PR #354, opened 2019-08-13
Use svgicon for scrolltrack

mhui:


wfro:
LGTM - more `Icon` => `SVGIcon` conversions are always welcome.

Let's test this one using `yarn link` (I can help with that, I haven't done it in a while myself so I need a refresher anyways) first since it's one of the more used components. I don't think there'll be any issues but can't hurt.

PR #358, opened 2019-08-30
Update eslint config and eslint to v6

nathanmarks:
Updated eslint and the config, and fixed some lint errors that the newer eslint version exposed.


src/styles/SetStyles.js
```diff
@@ -1,7 +1,8 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import Radium, { Style } from 'radium'
-import { Normalize, GlobalTheme } from '../index'
+import Normalize from './Normalize'
```
nathanmarks: circular dep

src/components/Typography/docs/helpers.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable react/jsx-no-undef */
+/* eslint-disable react/jsx-no-undef, react/prop-types */
```
nathanmarks: there's a component here with no prop type defs

src/utils/detectFeature.js
```diff
@@ -1,5 +1,12 @@
 import memoize from './memoize'
 
+export const isNodeEnv = () => typeof window === 'undefined'
```
nathanmarks: no-use-before-define

src/components/ScrollTrack/ScrollTrack.js
```diff
@@ -238,9 +238,9 @@ class ScrollTrack extends Component {
     } // already sliding
 
     const { parentWidth, trackWidth } = this.getNodeWidths()
+    const { onBeforeNext, scrollOffset } = this.props
```
nathanmarks: no-use-before-define

src/styles/SetStyles.js
dcocchia: oo tricky one

PR #361, opened 2019-09-03
Change circle caching

nathanmarks:
This is the correct way to cache yarn npm deps on circle, hopefully should fix the issue we had after the eslint upgrade from occurring again.


wfro:
Nice, thanks for fixing that.

PR #369, opened 2019-10-08
SelectionPills Component

loganmurdock:
## Business need
Support diet based filtering in search. Current navigation pills do not support multi-select, nor have the correct behavior from an accessibility standpoint. Request for a component that works as a filter checkbox with styling of a button:

https://instacart.atlassian.net/browse/CS-12631
https://instacart.quip.com/68zdARaoRKL3/2019Q3-Diet-Preferences-V2#eCfACAdxLT7

## SelectionPill
Individual selection pill with `isSelected`, `isDisabled` and `isFocused` states. `isFocused`  state is necessary to support outlines for keyboard users as a label based checkbox does not have native outline support.

selectionPill

## SelectionPills
Pills list component that controls the state of individual pills or allows for parent component controlled state. Select all button added when prop is included or allows for maximum selection count.

### Base behavior
selectionPills

### With select all button
withSelectAll

### With max selection count
selectionMax

### Browsers
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge via browserStack
- [ ] IE11 (unable due to incompatibility of styleguide with IE11. Opened [issue](https://github.com/instacart/Snacks/issues/370) to address)
- [x] Mobile safari on iPhoneX simulator
- [x] Mobile chrome on Android using browserStack

### A11y testing
- [x] VoiceOver on MacOS
- [x] Keyboard functionality and focus indicators
- [ ] Native client voiceover and rotor functionality


NinjaBanjo:
This looks incredibly close to `NavigationPills` https://instacart.github.io/Snacks/#navigationpills

they have an onClick, and isActive, can be used as siblings next to eachother. have a parent component that encompases multiple.

in addition, they almost look identical.

In what scenario would one use a `NavigationPill` and when would they use a `SelectionPill`?

loganmurdock:
@NinjaBanjo - as discussed directly through slack, the differences were significant due to the accessibility concerns that a selection filter creates, that I decided on a separate component. 

This component needs to be a checkbox rather than a link and has significantly different styling due to needing to style the label and hide the underlying checkbox. Additionally, it must hold a focus state in order to provide a focus outline for keyboard users. These constrains made it such that editing `NavigationPill` to meet these needs would require major re-working and cause the component to no longer be clean and readable (lots of conditionals based on what type was passed in). 

Ultimately the component seemed unique enough in functionality that it seemed warranted to write a separate set of components and test them separately.

src/components/SelectionPills/SelectionPill/SelectionPill.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  handleBlur = event => {
+    this.toggleFocus()
+    this.props.onBlur(event, { ...this.props, isFocused: !this.state.isFocused })
+  }
+
+  renderInputBtn({ selected }) {
+    const { aria, isDisabled, id } = this.props
+    const componentStyles = getStyles({})
```
wfro: small style nit - feels better to have `{}` set as the default argument for `getStyles`.

src/components/SelectionPills/SelectionPill/SelectionPill.js
```diff
@@ -0,0 +1,150 @@
...
+    const componentStyles = getStyles({
+      isSelected: selected,
+      isFocused,
+      isDisabled,
+      primaryForeground,
+      externalStyles: style,
+    })
+
+    return (
+      <li style={componentStyles.listElement} {...this.props.listElementAttributes}>
```
wfro: Let me know if `listElementAttributes` was for a specific reason, but this could be a good time to start the practice of spreading extra props to components without a separate prop like `elementAttributes`. Thinking something like 
```
const { id, snacksTheme, text, style, isDisabled, parentControlledState, ...other } = this.props

return (
      <li style={componentStyles.listElement} {...other}>
      ...
)
```

loganmurdock: Definitely no good reason for this. More following practice seen throughout snacks, but very supportive of moving towards spreading other props. Will update 👍

src/components/SelectionPills/SelectionPill/docs/SelectionPill.md
```diff
@@ -0,0 +1,38 @@
+**`SelectionPill` example:**
+
```
wfro: Below paragraph is good since it describes how to use the component, right above can you add a quick tl;dr for what `SelectionPill` is and maybe how it differs from navigation pills?

src/components/SelectionPills/__tests__/SelectionPills.spec.js
```diff
@@ -0,0 +1,162 @@
...
+    expect(bananasPill.checked).toBe(true)
+    expect(applePill.checked).toBe(false)
+
+    allPills.first().simulate('change')
+
+    expect(bananasPill.checked).toBe(false)
+    expect(applePill.checked).toBe(false)
+  })
+
+  xit('responds to parent controlled state correctly', () => {
```
wfro: stray xit

src/components/SelectionPills/__tests__/SelectionPills.spec.js
```diff
@@ -0,0 +1,162 @@
...
+    expect(onSelectPill.withArgs(3)).toBeTruthy()
+  })
+
+  it('disables remaining pills when max selected', () => {
+    const selectPills = [
+      { text: 'bananas', id: 'selection-1' },
+      { text: 'arugula', id: 'selection-2', isSelected: true },
+      { text: 'apple', id: 'selection-3' },
+    ]
+    const Pills = mount(
```
wfro: another style nit - we conventionally call this `wrapper` or `component` so it's not confused with a React element.

src/components/SelectionPills/__tests__/SelectionPills.spec.js
```diff
@@ -0,0 +1,162 @@
...
+import { StyleRoot } from 'radium'
+import { mount } from 'enzyme'
+import { spy } from 'sinon'
+import SelectionPills from '../SelectionPills'
+
+describe('SelectionPills', () => {
+  const testPills = [
+    {
+      text: 'bananas',
+      id: 'selection-1',
```
wfro: Are users always required to pass an explicit id for each pill?

src/components/SelectionPills/SelectionPills.js
```diff
@@ -0,0 +1,196 @@
...
+  // Disable not selected pills when max pills reached. Not applicable when parent controlled
+  isDisabledPill = pill => {
+    const { maxSelectionCount, parentControlledState } = this.props
+    const { selectedCount } = this.state
+    if (parentControlledState || pill.isSelected) return false
+    if (!selectedCount || !maxSelectionCount) return false
+    return selectedCount >= maxSelectionCount
+  }
+
+  renderSelectAllPill = () => {
```
wfro: More into design system territory, any thoughts on making the select all pill optional? This is something I'd use for a feature I'm working on, and there's a good chance I'd want to leave out a 'select all' option.

src/components/SelectionPills/SelectionPills.js
```diff
@@ -0,0 +1,196 @@
...
+    />
+  )
+
+  render() {
+    const { listAttributes, elementAttributes, parentControlledState, style } = this.props
+    const listToRender = parentControlledState ? this.props.pills : this.state.pillsList
+    const componentStyles = getStyles({ externalStyles: style })
+
+    return (
+      <ScrollTrack>
```
wfro: The only thoughts I have on the design are around `ScrollTrack`, is this something you're comfortable letting callers handle? e.g. I'm working on something `SelectionPills` would be great for, but mine are laid out as a grid vs. a scrollable list.

wfro:
Looks great, left some nits and a few questions around potential changes to make it slightly more flexible (select all and scroll track).

src/components/SelectionPills/SelectionPills.js
loganmurdock: The select all is totally optional and defaulted to false. If you look at the gifs I provided, the first gif shows the default behavior and does not include a select all. It's controlled through the `includeSelectAll` prop

src/components/SelectionPills/__tests__/SelectionPills.spec.js
loganmurdock: Currently, because I didn't want collisions and need to have a unique id to reference in the checkbox and label. I will update to use `text` and rely on the user to know to override when presenting duplicate text (for whatever reason).

src/components/SelectionPills/__tests__/SelectionPills.spec.js
NinjaBanjo: we should remove this test if it isn't going to run. is this unintentional?

src/components/SelectionPills/SelectionPill/SelectionPill.js
loganmurdock: Actually, doing this spreads a bunch of inappropriate objects on the list. In order to get around this, I would need to explicitly deconstruct all possible props even if they were not used here. It doesn't feel like the right treatment in this case.

mmarcuccio:
Awesome work! One important thing to add to the checklist is to test this component in the major browsers we support. Since this is a new component, it likely has a higher risk of cross-browser issues. Since we don't have any tooling set up around this, it would be a manual process for now (Browserstack might be the easiest way).

src/components/SelectionPills/SelectionPill/SelectionPill.d.ts
```diff
@@ -0,0 +1,50 @@
...
+  isDisabled?: boolean
+
+  /** Callback function called after pill click */
+  onClick(): void
+
+  /** Callback function called after pill gained focus */
+  onFocus(): void
+
+  /** Callback function called after pill has lost focus */
+  onBlur(): void
```
mmarcuccio: These three need more specific type signatures. The parameters are important for usage, and engineers will not be able to use these callbacks effectively from TS without defining types of the params here.

Also, if these are optional remember to use `?` to mark them as such.

src/components/SelectionPills/SelectionPill/SelectionPill.d.ts
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react'
+import { WithThemeInjectedProps, ApplyWithTheme } from '../../../styles/themer/withTheme'
+import { RadiumStyles, ElementAttributes } from '../../..'
+
+export interface SelectionPillProps extends WithThemeInjectedProps {
+  /** unique identifier for the pill */
+  id: string | number
```
mmarcuccio: Should this be `id?` or is the id required?

src/components/SelectionPills/SelectionPills.d.ts
```diff
@@ -0,0 +1,54 @@
...
+interface Pill
+  extends Pick<SelectionPillProps, 'text' | 'id' | 'isSelected' | 'isDisabled'> { }
+
+export interface SelectionPillsProps extends WithThemeInjectedProps {
+    /** Callback function called after pill click
+     * @param {SyntheticEvent} event The react `SyntheticEvent`
+     * @param {pill} object All pill attributes for pill selected
+     * @param {pillList} object List of all pills
+     */
+    onSelectPill(): void
```
mmarcuccio: Should be optional, and also needs the parameters to be defined here.

src/components/SelectionPills/SelectionPills.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    label: string
+
+    /** Option to only allow a maximum number of selected items. No restriction if not set. */
+    maxSelectionCount: number
+
+    /** Option to include a generated pill that will toggle all other pills on / off. Disabled if a maxSelectionCount or parentControlledState. */
+    includeSelectAll: boolean
+
+    /** Optional override of the select all pill label */
+    selectAllLabel: string
```
mmarcuccio: All of these should be optional.

src/components/SelectionPills/SelectionPills.js
```diff
@@ -0,0 +1,196 @@
...
+        parentControlledState
+      />
+    )
+  }
+
+  renderLabel = () => {
+    const { label, style } = this.props
+    if (!label) return null
+    return (
+      <span id="selectionPillsLabel" style={style.labelStyle}>
```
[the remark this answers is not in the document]
loganmurdock: Great callout! I was considering the impact of individual pills and their labels ids, but overlooked this point.

Fortunately this is used for associating the label for aria purposes, so I can update to use the label in the id. If there is a naming conflict it is because they have the same label and therefore same result for the screen reader.

src/components/SelectionPills/SelectionPill/SelectionPill.d.ts
```diff
@@ -22,16 +22,16 @@ export interface SelectionPillProps extends WithThemeInjectedProps {
...
-  onClick(): void
+  onClick?(e: React.MouseEvent<HTMLElement>, props: SelectionPillProps): void
 
   /** Callback function called after pill gained focus */
-  onFocus(): void
+  onFocus?(e: React.MouseEvent<HTMLElement>, props: SelectionPillProps, isFocused: boolean): void
 
   /** Callback function called after pill has lost focus */
-  onBlur(): void
+  onBlur?(e: React.MouseEvent<HTMLElement>, props: SelectionPillProps, isFocused: boolean): void
```
mmarcuccio: For these, I think props and focus are both properties in an object that is the second parameter.

src/utils/accessibility.js
```diff
@@ -0,0 +1,12 @@
+export const visuallyHidden = {
```
mmarcuccio: Since this is a constant, maybe we should use capital snake case like we do for things in `spacing`.

loganmurdock:
@mmarcuccio - Added testing section to the description 👍

PR #373, opened 2019-10-09
Add tooltip and dropdown style overrides

mmarcuccio:
This is mainly facilitate overriding `z-index` for these components as they currently use hard coded values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     },
     "setupFilesAfterEnv": [
       "<rootDir>/scripts/test/testSetup.js"
-    ]
+    ],
+    "snapshotSerializers": ["enzyme-to-json/serializer"]
   }
 }
--- a/src/components/Menus/DropdownMenu.js
+++ b/src/components/Menus/DropdownMenu.js
@@ -196,6 +196,8 @@ class DropdownMenu extends React.Component {
 
     const isOpen = this.state.open
 
+    console.log(isOpen)
+
     return (
       <div onKeyDown={this.handleKeyDown} style={style}>
         <div style={{ position: 'relative' }}>{this.renderTriggerElement()}</div>
@@ -206,6 +208,7 @@ class DropdownMenu extends React.Component {
             isOpen && [styles.menuContainerOpen, menuContainerOpenStyle],
             !isOpen && styles.menuContainerClosed,
           ]}
+          data-testid="menu-container"
         >
           <Slide in={isOpen} axis="y" style={styles.transitionContainer} offset={30}>
             <Fade in={isOpen} transitionTime={200}>
--- a/src/components/Menus/DropdownMenu.js
+++ b/src/components/Menus/DropdownMenu.js
@@ -56,6 +56,8 @@ class DropdownMenu extends React.Component {
     style: PropTypes.shape({}),
     /** Customize style of menu parent */
     menuContainerStyle: PropTypes.shape({}),
+    /** Customize style of menu parent when dropdown is open */
+    menuContainerOpenStyle: PropTypes.shape({}),
     /** Props passed to Menu component */
     menuProps: PropTypes.shape({}),
   }
@@ -190,7 +192,7 @@ class DropdownMenu extends React.Component {
   }
 
   render() {
-    const { style, children, menuProps, menuContainerStyle } = this.props
+    const { style, children, menuProps, menuContainerStyle, menuContainerOpenStyle } = this.props
 
     const isOpen = this.state.open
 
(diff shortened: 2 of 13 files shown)


src/components/Menus/__tests__/DropdownMenu.spec.js
```diff
@@ -34,6 +34,36 @@ it('renders DropdownMenu with icons and trigger correctly', () => {
...
+            leftIcon="twitterFilled"
+            leftIconStyles={{ color: '#1da1f2' }}
+          />
+        </DropdownMenu>
+      </div>
+    </StyleRoot>
+  )
+
+  wrapper.find('button').simulate('click')
+  expect(wrapper).toMatchSnapshot()
```
NinjaBanjo: the test says "renders dropdownmenu in open state correctly" 

i don't see where we validate that it's open. should we incluee that in this test? the snapshot alone feels somewhat useless. like we're just trying to get coverage.

src/components/Tooltip/__tests__/Tooltip.spec.js
```diff
@@ -40,8 +41,7 @@ describe('Tooltip', () => {
     const trigger = tooltip.find('button').last()
     trigger.last().simulate('click')
 
-    const tree = renderer.create(tooltip).toJSON()
-    expect(tree).toMatchSnapshot()
+    expect(tooltip).toMatchSnapshot()
```
NinjaBanjo: we should add some validation that the overlay actually appears. a snapshot doesn't provide that. it's not clear that the overlay is actually present in the dom

src/components/Menus/__tests__/DropdownMenu.spec.js
mmarcuccio: For dropdownmenu, open and close is only a difference in the value of z-inde on one of the child components, so it is captured in this snapshot test.  I can make a more explicit test that asserts the z-index value of that component.

src/components/Tooltip/docs/Tooltip.md
```diff
@@ -70,7 +70,7 @@ const toggleVisibility = () => {
 
     <div style={{padding: '30px', textAlign: 'center'}}>
       <Tooltip
-        target={state.trigger === 'icon' ? <Icon name='info' /> : 
+        target={state.trigger === 'icon' ? <Icon name='info' onClick={toggleVisibility} /> :
```
uorji3: Should we consider adding the `onClick` method to the entire Tooltip?

src/components/Menus/DropdownMenu.js
```diff
@@ -190,20 +192,23 @@ class DropdownMenu extends React.Component {
   }
 
   render() {
-    const { style, children, menuProps, menuContainerStyle } = this.props
+    const { style, children, menuProps, menuContainerStyle, menuContainerOpenStyle } = this.props
 
     const isOpen = this.state.open
 
+    console.log(isOpen)
```
uorji3: debug console log

jbender: Left in a debug

package.json
```diff
@@ -126,6 +126,7 @@
     },
     "setupFilesAfterEnv": [
       "<rootDir>/scripts/test/testSetup.js"
-    ]
+    ],
+    "snapshotSerializers": ["enzyme-to-json/serializer"]
```
wfro: This is already in `jest.config.js`, I'll make a note to delete this block. Were something around snapshots not working without this?

src/components/Menus/__tests__/DropdownMenu.spec.js
```diff
@@ -34,6 +34,43 @@ it('renders DropdownMenu with icons and trigger correctly', () => {
   expect(tree).toMatchSnapshot()
 })
 
+it('renders DropdownMenu in open state correctly', () => {
+  const menuContainerSelector = "div[data-testid='menu-container']"
```
wfro: There's a `getByTestId` helper you could use here(`import { getByTestId } from 'testing/utils'`). Though noticed you have the element type as part of the selector, let me know if `getByTestId` was missing something you needed and we can add it if so.

src/components/Tooltip/docs/Tooltip.md
mmarcuccio: In this demo, we want to hide/show the tooltip based on when the user clicks the icon or button. The wrapper includes other components that should not be click targets.

We could try to automatically pass any click handlers that are specified on the Tooltip to the Target property, but there is no guarantee it would accept an onClick handler.

When the tooltip is being used in the 'controlled input' mode as in this demo, we typically just leave full control to the engineer.

src/components/Menus/__tests__/DropdownMenu.spec.js
mmarcuccio: Made the tests more specific.

src/components/Tooltip/__tests__/Tooltip.spec.js
mmarcuccio: Made the tests more specific.

src/components/Menus/DropdownMenu.js
mmarcuccio: Thanks!

PR #374, opened 2019-10-10
[TextField] aria-required --> required

levkk:
I think `aria-required` does not dynamically update the field valid/invalid status when input is entered. `required` does that based on [MDN](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation#The_required_attribute) docs.


levkk:
CI failure because CI does not support forks I think

levkk:
Actually, that might break some of our stuff...

PR #385, opened 2019-12-06
Stephho/add expand icons

stephoho:
Screen Shot 2019-12-04 at 1 25 02 PM
This adds a new "expand" icon to support designs from Dan Shilov


loganmurdock:
This probably needs to be added to `ts` files as well: https://github.com/instacart/Snacks/blob/0be96fad6ead7fbb52df93b2cf83f822a9c8e4de/src/components/Icon/hexValues.d.ts
https://github.com/instacart/Snacks/blob/57bb6bcdc5afa054c751eab660e9f37ec0475170/src/components/SVGIcon/icons.d.ts

src/components/SVGIcon/icons/expand.svg
```diff
@@ -0,0 +1,6 @@
...
+<path fill-rule="evenodd" clip-rule="evenodd" d="M9.78667 13.5874C9.96045 13.4136 10.2422 13.4136 10.416 13.5874C10.5898 13.7612 10.5898 14.0429 10.416 14.2167L5.51733 19.1109H8.37689C8.49381 19.1114 8.60568 19.1586 8.68752 19.2422C8.76936 19.3257 8.81435 19.4385 8.81244 19.5554C8.81249 19.7974 8.61887 19.995 8.37689 19.9998H4.43556C4.32046 20.0001 4.21005 19.9543 4.12889 19.8727C4.04806 19.7912 4.00187 19.6817 4 19.5669V15.6265C4.00046 15.5096 4.04772 15.3977 4.13123 15.3159C4.21474 15.234 4.32754 15.189 4.44444 15.1909C4.68647 15.1909 4.88405 15.3845 4.88889 15.6265V18.4887L9.78667 13.5874Z" fill="#757575"/>
```
nathanmarks: we generally avoid hard-coded fills with fill rules and clip rules.. have we verified that this is usable in the same way as the other icons? 🤔

nathanmarks: https://github.com/instacart/Snacks/blob/master/src/components/SVGIcon/SVGIcon.js#L21

i don't think the `fill` passed here is going to work with these icons

stephoho: tbh, I just added the resources as handed over from design without checking-- I can't get the styleguide to compile locally. Any chance we can quickly pair on this so I know what to look for?

stephoho: Screen Shot 2019-12-09 at 10 00 43 AM

Updated!

PR #393, opened 2020-01-13
Bump webpack and remove from uglify => terser for minification

wfro:
Addresses part 1 of 2 of the CVE here https://github.com/instacart/snacks/network/alert/yarn.lock/serialize-javascript/open.

Bumps webpack to 4.41.5 (done with `yarn upgrade-interactive`).

The uglify plugin is deprecated, and the most recent version still
supports a version of serialize-javascript < 2.1.2.

TerserPlugin is the standard minification plugin for webpack these
days, and even has the same options API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "react-test-renderer": "16.4.2",
     "react-transition-group": "^2.2.1",
     "sinon": "^7.3.2",
+    "terser-webpack-plugin": "^2.3.2",
     "typescript": "^3.2.2",
     "uglifyjs-webpack-plugin": "^1.2.5",
     "url-parse": "1.4.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -80,7 +80,6 @@
     "react-transition-group": "^2.2.1",
     "sinon": "^7.3.2",
     "typescript": "^3.2.2",
-    "uglifyjs-webpack-plugin": "^1.2.5",
```
stefceror: Do we not need a new package for terser?

wfro: @stefceror Oh good catch, I wonder why this worked. It must have been able to resolve terser-webpack-plugin somehow, maybe since webpack has it as a dependency? I'll include it explicitly as a dependency though

stefceror: Thanks, I think it's best to be explicit about dependencies.

PR #394, opened 2020-01-13
Upgrade react + related deps; upgrade react-styleguidist

wfro:
Part 2 of 2 of resolving the CVE here: https://github.com/instacart/snacks/network/alert/yarn.lock/serialize-javascript/open

* Upgrades react/react-dom from 16.4.2 => 16.12.0
  * Also upgrades any deps that rely on certain versions of
    react like enzyme/react-test-renderer
* Upgrades react-styleguidist from 9 => 10. This was the original
  target package to upgrade since it had the outdated
  serialize-javascript dependency


package.json
```diff
@@ -73,10 +76,10 @@
     "prettier": "1.16.4",
     "prompt": "^1.0.0",
     "radium": "^0.22.0",
-    "react": "16.4.2",
-    "react-dom": "16.4.2",
-    "react-styleguidist": "9.0.9",
-    "react-test-renderer": "16.4.2",
+    "react": "^16.12.0",
```
NinjaBanjo: we should prob make sure this is a minor version change, as react updating is not insignificant (bundle sizes and all)

wfro: good idea 👍 , I think this would actually somehow be the first non-patch release.

wfro:
Had to rebase due to lockfile conflicts.

PR #420, opened 2020-05-14
tooltip: extend props to support backgroundColor

j4p3:
This PR extends Tooltip to accept `backgroundColor` style.

Use case:

https://instacart.quip.com/n7AzALPs30F7/PRD-Tooltips-1st-Journey-Education

The rest of the functionality described in the PRD can be accomplished with child components of the Tooltip.

An alternative approach would be to create a new `snacksStyle` preset to match whichever direction the onboarding tooltip design goes in the future. This seems like overkill for now, but is worth considering.

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -101,12 +101,6 @@ class Tooltip extends PureComponent {
       overlayStyle,
     } = this.props
 
-    if (style) {
-      console.warn(
-        'You have passed a `style` prop which will be deprecated in a future version of this component. Use `customStyle` instead.'
-      )
-    }
-
     return (
       <div>
         {this.renderTriggerElement()}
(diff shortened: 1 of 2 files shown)


j4p3:
👇  Codeclimate is correct about duplication, but there doesn't seem to be much we can do about it. Extracting PropTypes to their own object seems like a solution better suited to typescript.

src/components/Tooltip/Tooltip.js
```diff
@@ -103,7 +117,12 @@ class Tooltip extends PureComponent {
           onRootClose={this.handleHideToolTip}
           style={overlayStyle}
         >
-          <InnerToolTip size={size} style={style} arrowStyle={arrowStyle} snacksStyle={snacksStyle}>
+          <InnerToolTip
+            size={size}
+            customStyle={customStyle || style}
```
dmatteo, line 122: should we drop a comment here to remove the `|| style` once we remove support for `style`?

j4p3: `style` will be undefined here once we remove it from props, which should enforce its removal here. I'm fine to add @todos if you'd like, just didn't want to create too much mess in Snacks. Thoughts?

dmatteo: Sounds good 👍

PR #426, opened 2020-07-13
update text field to take an icon

JOlivier92:
allow our text field to take an optional icon prop. I added some default styling, and let the styling of the label and input up to the user. Added to snacks.io for documentation.

Screen Shot 2020-07-12 at 10 19 43 PM
Screen Shot 2020-07-12 at 10 24 47 PM


JOlivier92:
closing because the move to IDS is too close to justify updating this component.

PR #433, opened 2020-09-03
Add labelElementType prop to NavigationPills component to allow customization of tagName for component's label

glenndixon:
https://instacart.atlassian.net/browse/CS-22918

Re: the coveralls removal, I got direction from @nathanmarks to remove it on Slack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,3 +31,7 @@ jobs:
       - run:
           name: Upload code coverage report
           command: ./node_modules/.bin/codecov
+
+      - run:
+          name: Upload code coverage report - CoverAlls
+          command: ./node_modules/.bin/coveralls < ./coverage/lcov.info
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,3 @@ jobs:
       - run:
           name: Upload code coverage report
           command: ./node_modules/.bin/codecov
-
-      - run:
-          name: Upload code coverage report - CoverAlls
-          command: ./node_modules/.bin/coveralls < ./coverage/lcov.info
--- a/src/components/NavigationPills/NavigationPills.js
+++ b/src/components/NavigationPills/NavigationPills.js
@@ -70,8 +70,7 @@ class NavigationPills extends React.Component {
         return
       }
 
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -31,7 +31,3 @@ jobs:
       - run:
           name: Upload code coverage report
           command: ./node_modules/.bin/codecov
-
-      - run:
-          name: Upload code coverage report - CoverAlls
-          command: ./node_modules/.bin/coveralls < ./coverage/lcov.info
```
mmarcuccio, line 37: Was this an accidental inclusion?

src/components/NavigationPills/NavigationPills.js
```diff
@@ -70,8 +70,7 @@ class NavigationPills extends React.Component {
         return
       }
 
-      // eslint-disable-next-line jsx-a11y/label-has-associated-control
-      return <label style={styles.labelStyles}>{label}</label>
```
mmarcuccio, line 74: Instead of always using h2, we should probably make that a prop that can be passed in.

.circleci/config.yml
glenndixon: No.

src/components/NavigationPills/NavigationPills.js
glenndixon: Yeah I was thinking of adding something similar to the Text component's `elementType`. Here I might make it `labelElementType`.

I checked and this component is only used in two similar places in instacart (at the top of a retailer chooser). Do you think it's "worth it" to add this level of configurability rather than just making it an h2 and calling it a day?

mmarcuccio: I think it's worth the effort honestly. It also means you could make this change without impacting the other code by making the default be label.

glenndixon: Cool. Will update.

mmarcuccio (changes requested):
Make sure to update the type declaration for this component.

mmarcuccio (approved):
LGTM 👍

.circleci/config.yml
mmarcuccio: Looks like we are temporarily removing this and fixing CI after.

PR #434, opened 2020-09-03
A11y/checkbox focus

JOlivier92:
making checkboxes accessible by adding outlines to them when they are focused.


JOlivier92:
Screen Shot 2020-09-03 at 2 03 03 PM

PR #445, opened 2020-10-27
feat: add props id, aria-selected, aria-disabled, and aria-label to button component

dfadler1984:
I'm working on https://instacart.atlassian.net/browse/CS-26787 which will require me to add an id and a number of aria props to the button in https://github.com/instacart/carrot/blob/master/customers/instacart/app/assets/javascripts/store_app/components/shared/ServiceTypeControl/index.jsx#L26


dfadler1984:
Closing in favor of using `elementAttributes`

PR #450, opened 2021-03-25
rewrite themer to use a react context

NinjaBanjo:
This is a breaking change

consumers need to hook up the `ThemeProvider` to their own `Themer` class reference to get custom themes.

the default theme still works out of the box with no change


src/styles/themer/ThemerProvider.js
```diff
@@ -0,0 +1,42 @@
...
+  onThemeChange = () => {
+    const { tick } = this.state
+    // increment tick to force an update on the context
+    this.setState({ tick: tick + 1 })
+  }
+
+  render() {
+    const { themer, children } = this.props
+    const { tick } = this.state
+    return <ThemerContext.Provider value={{ themer, tick }}>{children}</ThemerContext.Provider>
```
nathanmarks, line 35: technically you don't need to pass `tick` here, because you're already creating a new object every time this `render()` method runs, which will happen after you `setState`. Context doesn't do deep comparisons to determine if the value has changed, it's referential.

dmatteo (dismissed):
LGTM tho I think Nathan is correct if you want to remove the tick trick (ha!)

dmatteo (approved):
Sweet!

NinjaBanjo:
published `0.0.157-prerelease.1` for now because of blocking pr, and this is p0

PR #460, opened 2021-07-27
scrollWidth causing incorrect calculatePosition

cliewdevelop:
proposing to change to offsetWidth

### DIFF
--- a/src/components/Tooltip/TooltipPosition.js
+++ b/src/components/Tooltip/TooltipPosition.js
@@ -53,7 +53,7 @@ class TooltipPosition extends PureComponent {
     const target = this.getTarget()
     const { overlay } = this
     const targetRect = this.getRect(target)
-    const documentWidth = document.documentElement.scrollWidth
+    const documentWidth = document.documentElement.offsetWidth
 
     const overlayRect = {
       top: targetRect.top,


cliewdevelop:
Screen Shot 2021-07-27 at 10 40 22 AM

Screen Shot 2021-07-27 at 10 40 28 AM

cliewdevelop:
When this.calculatePosition happens the first time, because we are using scrollWidth, the offset doesn't get calculate properly. The tooltip pops open to the right, and creates more space to the right, and auto scrolls there. (or it may not, depending on OS setting)

cliewdevelop:
Screen Shot 2021-07-27 at 10 42 29 AM

Screen Shot 2021-07-27 at 10 42 22 AM

*note: the orange blocks are intentionally hardcoded to visualize where the target dom for clicking is

cliewdevelop:
Afterwards, when scrollWidth becomes correct, this doesn't happen again. 
With the wrong scrollWidth, the calculation if block doesn't get triggered.

> else if (overlayDistanceFromRightEdge > documentWidth) {
https://github.com/instacart/Snacks/blob/master/src/components/Tooltip/TooltipPosition.js#L89


So far in debugging offsetWidth has been the correct width to use to trigger things

mmarcuccio (approved):

